PR #8, opened 2017-12-04
Smart contract local storage parser

heliorosa:
This PR adds a new method to the `*StateDB` type (package github.com/kowala-tech/kUSD/core/state): `UnmarshalState(addr *common.Address, first common.Hash, r interface{}) error`

The method `UnmarshalState` can also unmarshal custom types that implement `StateUnmarshaler` interface.

Solidity mappings must be unmarshaled using the provided type: `Mapping`


core/state/statedb.go
```diff
@@ -233,10 +233,11 @@ func (self *StateDB) GetCodeHash(addr common.Address) common.Hash {
 
 func (self *StateDB) GetState(a common.Address, b common.Hash) common.Hash {
```
yourheropaul: The idiomatic way of doing this in Go is by naming the return variables. Like,

```
func (self *StateDB) GetState(a common.Address, b common.Hash) (r common.Hash) {
    // ...
    return
}
```

That way you don't have to duplicate the declaration.

PR #29, opened 2018-01-19
Rewarded clique

yourheropaul:


yourheropaul:
I think the block reward implementation is pretty much exactly what's in the whitepaper, and the Clique approach is really neat.  Looks good to me! 

We need to decide where the transaction fees go, so I'll create a separate issue for that.

PR #49, opened 2018-01-29
Dockerfiles for the kusd and bootnode

acroca:
Creates two dockerfiles, one for the bootnode and one for the kusd.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -22,3 +22,21 @@ pipeline:
       - crc32
     when:
       event: tag
+
+  docker_kusd:
+    image: plugins/docker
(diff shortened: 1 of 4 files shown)


yourheropaul:
The Docker setup looks great! For Docker Hub, we don't need a push (I don't think) because we can set up automated builds in hub.docker.com. That system will poll the repo for changes and create versions for each tag. We can make stable, dev and feature branch versions too.

acroca:
I thought of using the CI to make sure the builds are always based on green branches, so if master or development is broken it won't get push to docker.

Issue #59, opened 2018-02-02
[ethereum] import/review package tests


rgeraldes:
Base: 1.73

PR #61, opened 2018-02-05
Fix rebase issues

acroca:
I've found some issues with the recent rebase:
- Some imports were using `ethereum/go-ethereum` instead of `kowala-tech/kUSD`
- The state runner has dependencies with `/tests`, package we didn't import. It seems unnecessary for us, so I removed it.
- Some code in a simulated backend was wrong.


acroca:
@rgeraldes is it necessary? The state runner has a dependency with some test code

rgeraldes:
I'm not 100 % sure and that's why I think we should comment the code + add a note (@TODO).

acroca:
We can also add a dummy implementation of the dependency from tests, so the staterunner doesn't change and as soon as #59 is done it'll all continue working. Makes sense?

rgeraldes:
That works too!

acroca:
Done, added a fake implementation of the dependencies, just to let it compile.

PR #66, opened 2018-02-06
validator: proposer selection

### LINKED ISSUE #32: [consensus] Complete the proposal mechanism.
- The current model selects the first validator as the proposer.

rgeraldes:
- Included a field called weight to the validator.
- priority queue to identify the proposer (max weight)
- The contract includes a summary to helps us identify if the validator set changed.

Closes #32 

Notes:
- The genesis validators need to use full sync (due to this condition: https://github.com/ethereum/go-ethereum/blob/master/eth/handler.go#L172 - fast sync never gets disabled if the peer does not connect to a peer with a block number bigger than the current one, which will happen for a genesis validator)


rgeraldes:
Noticed just now (reviewing) that if the validator set does not change, it's cheaper, to do a peek operation and update just the proposer weight, and use the same priority queue for the next round, instead of create a new priority queue on each election.

rgeraldes:
ready for review @yourheropaul @acroca

contracts/network/contracts/network.sol
```diff
@@ -23,6 +23,8 @@ contract Network {
     // minimum deposit value to participate in the consensus
     uint public minDeposit = 100000;
 
+    bytes32 public votersSummary;
```
acroca: I find the variable name confusing, had to check the implementation to find out it's more like a checksum. Maybe can be called `votersChecksum` or something more descriptive?.

core/types/validator.go
```diff
@@ -1,53 +1,89 @@
...
 func (set *ValidatorSet) Proposer() common.Address {
-	// @TODO (rgeraldes) complete - return the first validator for now
-	return set.validators[0].Address()
+	return set.proposer.address
 }
+
+func (set *ValidatorSet) Contains(addr common.Address) bool {
+	_, ok := set.membership[addr]
+	return ok
+}
```
acroca: It is recommended to have a new line at the end of the files. Most editors can do that for you, vscode does it with `"files.insertFinalNewline": true`.

kusd/validator/validator.go
```diff
@@ -259,17 +259,66 @@ func (val *Validator) restoreLastCommit() {
...
+					old := val.validators.Get(validator.Addr)
+					weight = old.Weight()
+				} else {
+					// new validator
+					weight = big.NewInt(0)
+				}
+				validators[i] = types.NewValidator(validator.Addr, validator.Deposit.Uint64(), weight)
+			}
+			val.validators = types.NewValidatorSet(validators)
+			val.validatorsSummary = summary
```
acroca: This whole code (from the `GetVoterCount...`) it's exactly the same as the previous code except for the weight. I think this version can be used in both places because it will never find an `old` validator, so it'll use `0` weight in all cases, right? If so, this could be refactored to something like `val.buildValidators()`.

contracts/network/contracts/network.sol
rgeraldes: I agree, I had the same feeling about the name.

core/types/validator.go
rgeraldes: For some reason my vs code extension is crashing a lot of times lately and that's the reason why the line was not added - by default it adds a line to the end of the file.

PR #71, opened 2018-02-09
tests: Console & Faucet

### LINKED ISSUE #69: [testnet] test console & faucet

rgeraldes:
This PR covers the issues found during the faucet and console tests. Closes #69 

**Context:** 
- for the eth module APIs the flow is
console call > execute js in jsre (web3 call) > rpc client > rpc server

**Important changes:**
- replaced the 'kusd' namespace with 'eth' (1)
- replaced the faucet html & binding with a new version (vs code added garbage - related to js formats - while replacing mUSD with kUSD)
- tested successfully the validator api

**Important Notes:**
- Modifications to the api of the core protocol won't work for now for console calls.
Ex: getEtherbase vs getCoinbase because the calls go through web3 which has not been changed yet. (Important: this is the only modification).
- The  #Faucet's network id needs to match the chain otherwise the handshake will fail
- Make sure that faucet account is pre funded. (The puppeth default value is enough)

** Very very important **
- Make sure that the initial balance of the network contract is equal to the sum of all the deposits of the genesis validators.

**Tests**

Faucet 

- Transaction:

>INFO [02-12|12:24:11] Imported new chain segment               blocks=1 txs=1 mgas=021 elapsed=5.053ms   mgasps=4.155 number=13 hash=a4fc6b…292bee

> eth.getBalance("0xcfff0fdae894be2ed95e02f514b3fbfc1bf41656")
1000000000000000000

Console:

- Transaction:

> eth.getBalance("0xcfff0fdae894be2ed95e02f514b3fbfc1bf41656")
0

> eth.sendTransaction({from:eth.coinbase, to:"0xcfff0fdae894be2ed95e02f514b3fbfc1bf41656", value: web3.toWei(0.05, "ether")})
INFO [02-12|12:56:57] Submitted transaction                    fullhash=0xbd195474fe47a57f8ce7972c35ee83debb9657dcad198325927c1e7f94547a63 recipient=0xCFff0fdae894Be2eD95E02f514b3fbFc1Bf41656
"0xbd195474fe47a57f8ce7972c35ee83debb9657dcad198325927c1e7f94547a63"

> eth.getBalance("0xcfff0fdae894be2ed95e02f514b3fbfc1bf41656")
50000000000000000

(1)
"in short we are using at the moment the 'kusd' namespace which appears as a module in the console, but the internals of the console rely on var eth = web3.eth. A new alias for kusd is not enough as they have some conditions in place and also calls to eth (core protocol) go through web3 which is using eth hardcoded as the namespace and we would have to hardcode a mapping from eth to kusd on the rpc server. I think that the easiest solution would be to either have eth as the namespace, but the negative about this is that in the console we will see eth instead of kusd. The positive thing about it is that every ethereum dapp would be compatible right away. The other option is to modify web3 to use our namespace. Ethereum dapps just would have to change var eth = web3.eth to var eth = web3.kusd."


kusd/validator/states.go
```diff
@@ -69,6 +69,16 @@ func (val *Validator) notLoggedInState() stateFn {
 			}
 		}
 	} else {
+		// sanity check
+		isVoter, err := val.network.IsVoter(&bind.CallOpts{}, val.account.Address)
```
jmatosp: this is the same code as above, if we exit on err always  (log.Crit?) maybe is worth moving this into a assertIsVoter/assertGetIsVoter (or similar) method helper returning only a bool as assert implies failure/panic. 

this methos is long already, is the block above a makeDeposit()?

I feel like i've done something today :)

acroca:
Regarding the namespace `kusd->eth`, we could also keep `eth` as it is and create a `kusd` var that contains everything from `eth` plus our extra code. Would that make sense, @rgeraldes?

kusd/validator/states.go
```diff
@@ -69,6 +69,16 @@ func (val *Validator) notLoggedInState() stateFn {
...
+		if err != nil {
+			log.Crit("Failed to verify the voter information", "err", err)
+			return nil
+		}
+		if !isVoter {
+			log.Crit("Invalid genesis - genesis validator needs to be registered as a voter", "address", val.account.Address)
```
acroca: If it's not a voter, can still start validation at this point? I understand this branch of the code is for genesis validators that start after genesis validation (block > 0) and are not voters.

kusd/validator/states.go
```diff
@@ -236,7 +246,7 @@ func (val *Validator) commitState() stateFn {
 	voter, err := val.network.IsVoter(&bind.CallOpts{}, val.account.Address)
 	if err != nil {
 		// @TODO (rgeraldes) - complete
-		//log.Error()
+		log.Crit("Failed to verify if the validator is a voter")
```
acroca: Should we `return nil` here?

rgeraldes:
@acroca on the kusd->eth, I've tried that approach (creating an alias) but they have conditions in place that are not compatible with it and they map the rpc calls manually. https://github.com/ethereum/web3.js/blob/develop/lib/web3/methods/eth.js#L323.
Also ethereum provides costumizations over the web3 here https://github.com/ethereum/go-ethereum/blob/master/internal/web3ext/web3ext.go and once again it uses eth hardcoded.
The code is a big spaghetti :) I discussed this topic with @yourheropaul some days ago and he decided to leave as is for now. (possible multiple coins for kowala)

kusd/validator/states.go
rgeraldes: All the validators that are not genesis, are not voters. They will enter the other condition and are forced to make a deposit. In the future we need to consider additional scenarios such as if a process crashes. (Related to the punishment issue)

kusd/validator/states.go
rgeraldes: We need to study these situations carefully (related to the punishment issue).

PR #72, opened 2018-02-13
Running test packages on build

jmatosp:
Added Test run on feature push
Added fmt and cyclo complexity check
Added a few simple test to monitor CI results

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,13 +1,26 @@
+workspace:
+  base: /go
+  path: src/github.com/kowala-tech/kUSD
+
 pipeline:
 
-  test-and-build:
+  test:
     image: kowalatech/go:1.0.1
+    commands:
+      - make test
+      - make lint
+    when:
+      event: [push]
+      branch: [master, dev, feature/*, hotfix/*]
+
+  build:
+    image: golang
     commands:
       - make
     when:
       event: [push, tag]
       branch: [master, dev, refs/tags/*]
-      
+
   github_release:
     image: plugins/github-release
     secrets: [ github_release_api_key ] 
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,9 @@ ios:
 test: all
 	build/env.sh go run build/ci.go test
 
+lint: all
+	build/env.sh go run build/ci.go lint
+
 clean:
 	rm -fr build/_workspace/pkg/ $(GOBIN)/*
 
--- a/build/.gometalinter.json
+++ b/build/.gometalinter.json
@@ -1,0 +1,19 @@
+{
+  "Vendor": true,
+  "Deadline": "2m",
+  "Sort": ["linter", "severity", "path", "line"],
+  "EnableGC": true,
+  "WarnUnmatchedDirective": true,
+  "DisableAll": true,
+  "Enable": [
+    "deadcode",
+    "gocyclo",
+    "gofmt",
+    "misspell",
+    "unconvert",
+    "unused",
+    "vet"
+  ],
+  "Cyclo": 13,
+  "LineLength": 100
+}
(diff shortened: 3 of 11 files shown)


.drone.yml
```diff
@@ -1,13 +1,36 @@
+workspace:
+  base: /go
+  path: src/github.com/kowala-tech/kUSD
+
 pipeline:
 
-  test-and-build:
+  test:
+    image: golang:1.9
+    commands:
```
yourheropaul: Can all this be a Makefile task so it's reproducible locally?

.drone.yml
```diff
@@ -1,13 +1,36 @@
+workspace:
+  base: /go
+  path: src/github.com/kowala-tech/kUSD
+
 pipeline:
 
-  test-and-build:
+  test:
+    image: golang:1.9
```
yourheropaul: This is a prime candidate for a standardised tooling and testing image. We have https://hub.docker.com/r/kowalatech/go (currently serving version 1.9.2 with AWS tools and supervisord) that could easily be extended to include gometalinter and other useful packages by default. 

We already use it for local development and testing of our closed source packages (see https://github.com/kowala-tech/core) and it would be awesome to have a standard base image for all Kowala Go work.

.drone.yml
yourheropaul: (by 'all this' I mean the commands that run the tests)

jmatosp: yes, makes sense, didn't realise our pipeline was well setup

.drone.yml
jmatosp: cool, I will add the deps to docker dev image.

build/.gometalinter.json
```diff
@@ -0,0 +1,19 @@
...
+    "gocyclo",
+    "gofmt",
+    "misspell",
+    "unconvert",
+    "unused",
+    "vet"
+  ],
+  "Cyclo": 13,
+  "LineLength": 100
+}
```
acroca: It's recommended to have an empty line at the end of all files

.drone.yml
```diff
@@ -1,13 +1,25 @@
+workspace:
+  base: /go
+  path: src/github.com/kowala-tech/kUSD
+
 pipeline:
 
-  test-and-build:
+  test:
+    image: kowalatech/go:1.0.1
```
acroca: Why not `golang:1.9.2-alpine`? I think the only difference is that ours have supervisor, but I don't think we need it here

build/ci.go
```diff
@@ -122,6 +122,15 @@ var (
 	// Note: wily is unsupported because it was officially deprecated on lanchpad.
 	// Note: yakkety is unsupported because it was officially deprecated on lanchpad.
 	debDistros = []string{"trusty", "xenial", "zesty", "artful"}
+
+	// stable is a list of packages known to be stable to do full test, list and cyclo
```
acroca: comment mentions `stable` instead of `stablePackages`

build/.gometalinter.json
jmatosp: :+1:

.drone.yml
jmatosp: your right, don't need that gometalinter from base box

build/ci.go
jmatosp: :+1:

Makefile
```diff
@@ -44,9 +44,12 @@ ios:
 	@echo "Done building."
 	@echo "Import \"$(GOBIN)/Kusd.framework\" to use the library."
 
-test: all
+test: all lint
```
rgeraldes: I think that test should not include lint - different areas.
ex: you have a local branch that needs to be tested several times but you're not interested in running the lint tools until the implementation is mature.
what do you think about a different tag like this one > ci: lint mock test-unit test-e2e

jmatosp: yes, we can do that, agree. but we need to run this always on build to ensure.

also I think in day-to-day coding you will need to run these often otherwise you will have lots of work (fix lint) to do on a mature feature

PR #77, opened 2018-02-14
Validator updates

### LINKED ISSUE #73: [validator] transaction uses hardcoded gas prices
### _Type_:
Issue

The transactions made by the validator use hardcoded gas prices.
We need to replace the hardcoded values with good estimations.
We may need to dynamically change the gas price if the estimation is not accurate. (research)

### _Priority_:
Low

rgeraldes:
Important changes:
- validator is only stopped if its validating.
- transactions made by the validator are now using the core estimations.
- If the validator's deposit is not enough, the validation does not start, but the node continues its activity as an archival node because we need to take the console into account - he can join again as soon as he wants (and if he has the deposit).
- The validator weight will be set to 0 every-time a new validator joins the elections - the validator that joins a election is not aware of the previous weights of other validators and those will need to be shared. (I will create an issue to address this issue/research the topic)
- A validator can start the election alone (relies on a forced sync interval) - If the peer does not get to other peers in 10 seconds there's a forced sync which will result in enabling the start of the consensus. Note that 10 seconds is more than enough to have feedback from other peers (in case there are peers)

Closes #73


kusd/validator/utils.go
```diff
@@ -0,0 +1,51 @@
...
+	"fmt"
+	"math/big"
+
+	kowala "github.com/kowala-tech/kUSD"
+	"github.com/kowala-tech/kUSD/accounts/abi/bind"
+	"github.com/kowala-tech/kUSD/common"
+	"github.com/kowala-tech/kUSD/core/types"
+)
+
+func getTransactionOpts(contractBackend bind.ContractBackend, from common.Address, value *big.Int, chainID *big.Int) (*bind.TransactOpts, error) {
```
jmatosp: this is a good place to test.
the func is loo large, you could split the anonymous func and test both in isolation

PR #88, opened 2018-02-15
refactored func into smaller bits, PR TO TESTS-CI

jmatosp:
Refactoring some functions on Consensus for reducing complexity

Important changes:

Closes

### DIFF
--- a/kusd/validator/states.go
+++ b/kusd/validator/states.go
@@ -16,12 +16,11 @@ import (
 // @TODO (rgeraldes) - confirm
 // work is the proposer current environment and holds all of the current state information
 type work struct {
-	state     *state.StateDB
-	header    *types.Header
-	tcount    int
-	failedTxs types.Transactions
-	txs       []*types.Transaction
-	receipts  []*types.Receipt
+	state    *state.StateDB
+	header   *types.Header
+	tcount   int
+	txs      []*types.Transaction
+	receipts []*types.Receipt
 }
 
 // stateFn represents a state function
@@ -135,7 +134,7 @@ func (val *Validator) newRoundState() stateFn {
 }
 
 func (val *Validator) newProposalState() stateFn {
-	timeout := time.Duration(params.ProposeDuration+uint64(val.round)*params.ProposeDeltaDuration) * time.Millisecond
+	timeout := time.Duration(params.ProposeDuration+val.round*params.ProposeDeltaDuration) * time.Millisecond
 
 	if val.isProposer() {
 		log.Info("Proposing a new block")
@@ -163,7 +162,7 @@ func (val *Validator) preVoteState() stateFn {
 
 func (val *Validator) preVoteWaitState() stateFn {
 	log.Info("Waiting for a majority in the pre-vote sub-election")
-	timeout := time.Duration(params.PreVoteDuration+uint64(val.round)*params.PreVoteDeltaDuration) * time.Millisecond
+	timeout := time.Duration(params.PreVoteDuration+val.round*params.PreVoteDeltaDuration) * time.Millisecond
 
 	select {
 	case <-val.majority.Chan():
@@ -184,7 +183,7 @@ func (val *Validator) preCommitState() stateFn {
 
 func (val *Validator) preCommitWaitState() stateFn {
 	log.Info("Waiting for a majority in the pre-commit sub-election")
-	timeout := time.Duration(params.PreCommitDuration+uint64(val.round)+params.PreCommitDeltaDuration) * time.Millisecond
+	timeout := time.Duration(params.PreCommitDuration+val.round+params.PreCommitDeltaDuration) * time.Millisecond
 	// @TODO (rgeraldes) - move to a post processor state
 	defer val.majority.Unsubscribe()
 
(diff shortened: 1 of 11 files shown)


kusd/validator/states.go
```diff
@@ -31,54 +31,22 @@ func (val *Validator) notLoggedInState() stateFn {
...
 	// @NOTE (rgeraldes) - sync was already done at this point and by default the investors will be
 	// part of the initial set of validators - no need to make a deposit if the block number is 0
 	// since these validators will be marked as voters from the start
-	if !isGenesis || (isGenesis && val.chain.CurrentBlock().NumberU64() > 0) {
-		// Subscribe events from blockchain
-		chainHeadCh := make(chan core.ChainHeadEvent)
-		chainHeadSub := val.chain.SubscribeChainHeadEvent(chainHeadCh)
-		defer chainHeadSub.Unsubscribe()
-
+	if isFirstBlock(isGenesis, val) {
```
rgeraldes: isFirstBlock should be used only for this part of the condition - val.chain.CurrentBlock().NumberU64() > 0
Genesis validators are election voters (consensus) that are included from the block 0 (genesis block) - could be an assert like isVoter.

jmatosp: ha!

kusd/validator/states.go
```diff
@@ -91,6 +52,60 @@ func (val *Validator) notLoggedInState() stateFn {
...
+
+func (val *Validator) isGenesisRejoining(isGenesis bool) bool {
+	return isGenesis && val.isNotFirstBlock()
+}
+
+func (val *Validator) isFirstBlock() bool {
+	return val.chain.CurrentBlock().NumberU64() == 0
+}
+
+func (val *Validator) isNotFirstBlock() bool {
```
acroca: I think using a `!` to negate the `isFirstBlock` looks better. If in the future some is refactoring a code where at first it was using `isNotFirstBlock` and the new logic requires checking for first block, I doubt that person will think there's another function with the negative and it'd use `!val.isNotFirstBlock()`. It'd happen to me :P

kusd/validator/states.go
```diff
@@ -91,6 +52,60 @@ func (val *Validator) notLoggedInState() stateFn {
...
+	for {
+		select {
+		case _, ok := <-chainHeadCh:
+			if !ok {
+				log.Info("Failed reading on Header channel")
+			}
+
+			isVoter := val.assertGetIsVoter()
+			if isVoter {
+				break L
```
acroca: I think now we can just `return` and not use labels.

kusd/validator/states.go
```diff
@@ -91,6 +52,60 @@ func (val *Validator) notLoggedInState() stateFn {
...
+			}
+		}
+	}
+}
+
+func (val *Validator) assertGetIsVoter() bool {
```
acroca: What's the difference between `assert...` and `is...`? both return a bool

kusd/validator/states.go
```diff
@@ -104,7 +119,7 @@ func (val *Validator) newElectionState() stateFn {
 	// @NOTE (rgeraldes) - wait for txs - sync genesis validators, round zero for the first block only.
 	if val.blockNumber.Cmp(big.NewInt(1)) == 0 {
 		numTxs, _ := val.backend.TxPool().Stats() //
-		if val.round == 0 && numTxs == 0 {        //!cs.needProofBlock(height)
+		if val.round == 0 && numTxs == 0 { //!cs.needProofBlock(height)
```
acroca: if you change the aligned comment I think you can remove the one from the previous line 😃

kusd/validator/states.go
jmatosp: this is from clean code, readability first. but I'm not too fussed about it. They say use the one that is most common (usage)

kusd/validator/states.go
jmatosp: It's been year's since I've seen Gotos :) happy you guys wanna ditch that. I didn't here, because didn't went thru the logic, was just splinting in smaller chunks

kusd/validator/states.go
jmatosp: assert as in fail otherwise, there is a log.Crit that panic/exits so this func implies get or fail. It's a hack as I think most errors should be recoverable, or at least managed and bubbled up

kusd/validator/states.go
jmatosp: :+1: this was a refactor using my ide tooling, and fmted after

jmatosp: actually, there should be NO commented code

kusd/validator/states.go
acroca: I see. Then, all `is...` functions should have the `isNot...` as well? It's not something very common, I've only seen this kind of things in testing libraries for assertions, but not for general code.

kusd/validator/states.go
acroca: I see, didn't realise the `Crit` panics. Thanks!

kusd/validator/states.go
jmatosp: Not necessarily all functions should have the Not,
The main point is readability, in some places the ! is easy to misses at first look
I'm being a bit nitpicking :) I will remove one

PR #95, opened 2018-02-19
Feature/sync waiter - Step 1

jmatosp:
A function that waits for Block Sync to finish and then calls a provided function
Example:
```
mux := new(event.TypeMux)
defer mux.Stop()
SyncWaiter(
   mux,
   func() { fmt.Print("synchronization done!) },
   func() { fmt.Print("synchronization failed!") },
)
```

### DIFF
--- a/kusd/validator/sync_waiter.go
+++ b/kusd/validator/sync_waiter.go
@@ -1,0 +1,30 @@
+package validator
+
+import (
+	"errors"
+	"github.com/kowala-tech/kUSD/event"
+	"github.com/kowala-tech/kUSD/kusd/downloader"
+)
+
+var (
+	errFailedEventReceived      = errors.New("FailedEvent received")
+	errFailedToReceiveDoneEvent = errors.New("failed to receive DoneEvent")
+)
+
+// SyncWaiter subscribes to DoneEvent and FailedEvent on a event.TypeMux
+// and when one of those events is received call a user function.
+// Example usage to delay Validator start after block sync has finished
+func SyncWaiter(eventMux *event.TypeMux) error {
+	events := eventMux.Subscribe(downloader.DoneEvent{}, downloader.FailedEvent{})
+	defer events.Unsubscribe()
+	for ev := range events.Chan() {
+		switch ev.Data.(type) {
+		case downloader.DoneEvent:
+			return nil
+		case downloader.FailedEvent:
+			return errFailedEventReceived
+		}
+	}
+
+	return errFailedToReceiveDoneEvent
+}
--- a/kusd/validator/sync_waiter_test.go
+++ b/kusd/validator/sync_waiter_test.go
@@ -1,0 +1,49 @@
+package validator
+
+import (
+	"github.com/kowala-tech/kUSD/event"
+	"github.com/kowala-tech/kUSD/kusd/downloader"
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+func TestSyncWaiterNoError(t *testing.T) {
+	mux := new(event.TypeMux)
+	defer mux.Stop()
+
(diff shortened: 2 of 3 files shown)


jmatosp:
@rgeraldes 
do we need to subscribe to StartEvent?
do we callback on FailedEvent?

rgeraldes:
**Start Event**
We don't need it. I noticed that some weeks ago but forgot to open an issue.

**Failed Event**
Ethereum handles the FailedEvent the same way as DoneEvent. I had a look at the code and the event is thrown when err != nil while trying to sync with a peer. In Ethereum's case it's ok to have it fail because as soon as the node connect to other peer the challenge is stopped and he tries to sync. In our case we should log the error. We should just move forward with a done event.

rgeraldes:
A note of warning for future tests that rely on times: There's a window of 10 seconds to connect to peers otherwise there will be a forced sync for the best peer (which will be nil if there are no peers) and the node will consider himself synced - use case for the genesis validator.
Duration: https://github.com/kowala-tech/kUSD/blob/master/kusd/sync.go#L16
Note also that 10 seconds is a lot of time and that as soon as a normal node connects he will receive enodes right away from the bootnode, as long as there are other peers.

jmatosp:
@rgeraldes I noticed a potential bug when testing this. the Validator subscribes events DoneEvent and FailedEvent, but it could occur that when it starts, those events have already ocurred, is this a possible scenario? On tests it is, that's why I added a delay

kusd/validator/sync_waiter_test.go
```diff
@@ -0,0 +1,85 @@
...
+		func() { assert.Fail(t, "function was called") },
+		func() { assert.Fail(t, "function was called") },
+	)
+
+	// give a chance for go SyncWaiter to subscribe to channel before POST
+	time.Sleep(time.Millisecond * 10)
+	err := mux.Post(downloader.StartEvent{})
+	assert.NoError(t, err, "error posting event")
+
+	// wait a but, NO function should be called
```
rgeraldes: but > bit

kusd/validator/sync_waiter.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/kowala-tech/kUSD/event"
+	"github.com/kowala-tech/kUSD/kusd/downloader"
+)
+
+// SyncWaiter subscribes to DoneEvent and FailedEvent on a event.TypeMux
+// and when one of those events is received call a user function.
+// Example usage to delay Validator start after block sync has finished
+func SyncWaiter(eventMux *event.TypeMux, callback func(), callbackError func()) {
+	events := eventMux.Subscribe(downloader.DoneEvent{}, downloader.FailedEvent{})
+	defer events.Unsubscribe()
```
acroca: Wouldn't it be better to unsubscribe before calling the callbacks and not after? We don't know how much time will the callback take, so I feel it's a bit more efficient to unsubscribe before calling them

jmatosp: agree, wasn't sure, this was previous behaviour.

kusd/validator/sync_waiter.go
```diff
@@ -0,0 +1,24 @@
...
+// Example usage to delay Validator start after block sync has finished
+func SyncWaiter(eventMux *event.TypeMux) error {
+	events := eventMux.Subscribe(downloader.DoneEvent{}, downloader.FailedEvent{})
+	defer events.Unsubscribe()
+	for ev := range events.Chan() {
+		switch ev.Data.(type) {
+		case downloader.DoneEvent:
+			return nil
+		case downloader.FailedEvent:
+			return errors.New("FailedEvent received")
```
acroca: You can move this error to a constant so it can be referenced.

PR #97, opened 2018-02-19
Refactored validator to interface

jmatosp:
Added Validator interface with all methods from validator struct
Added return errors in methods to gracefully handle them
Logic found in handler when receiving messages (validator.Validating) is now in the Validator it self, and an error is raised if not validating (allows to handle the error), please verify this one as is a slight change in program flow (handler.go)

Important changes:
now most methods might return an identifiable error

### DIFF
--- a/kusd/handler.go
+++ b/kusd/handler.go
@@ -61,7 +61,7 @@ type ProtocolManager struct {
 
 	downloader *downloader.Downloader
 	fetcher    *fetcher.Fetcher
-	validator  *validator.Validator
+	validator  validator.Validator
 	peers      *peerSet
 
 	SubProtocols []p2p.Protocol
@@ -85,7 +85,7 @@ type ProtocolManager struct {
 
 // NewProtocolManager returns a new kowala sub protocol manager. The Kowala sub protocol manages peers capable
 // with the kowala network.
-func NewProtocolManager(config *params.ChainConfig, mode downloader.SyncMode, networkID uint64, mux *event.TypeMux, txpool txPool, engine consensus.Engine, blockchain *core.BlockChain, chaindb kusddb.Database, validator *validator.Validator) (*ProtocolManager, error) {
+func NewProtocolManager(config *params.ChainConfig, mode downloader.SyncMode, networkID uint64, mux *event.TypeMux, txpool txPool, engine consensus.Engine, blockchain *core.BlockChain, chaindb kusddb.Database, validator validator.Validator) (*ProtocolManager, error) {
 	// Create the protocol manager with the base fields
 	manager := &ProtocolManager{
 		networkID:   networkID,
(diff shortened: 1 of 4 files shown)


kusd/handler.go
```diff
@@ -628,48 +628,36 @@ func (pm *ProtocolManager) handleMsg(p *peer) error {
...
 		var proposal types.Proposal
 		if err := msg.Decode(&proposal); err != nil {
 			return errResp(ErrDecode, "msg %v: %v", msg, err)
 		}
-		pm.validator.AddProposal(&proposal)
-
-	case msg.Code == VoteMsg:
-		// @TODO (rgeraldes) - review flow (we will not need this condition)
-		if !pm.validator.Validating() {
+		if err := pm.validator.AddProposal(&proposal); err != nil {
```
acroca: Should we add some log if this returns an `err`?

kusd/handler.go
```diff
@@ -628,48 +628,36 @@ func (pm *ProtocolManager) handleMsg(p *peer) error {
...
 			return errResp(ErrDecode, "msg %v: %v", msg, err)
 		}
-
-		p.MarkVote(vote.Hash())
-		pm.validator.AddVote(&vote)
-
-	case msg.Code == BlockFragmentMsg:
-		// @TODO (rgeraldes) - review flow (we will not need this condition)
-		if !pm.validator.Validating() {
+		if err := pm.validator.AddVote(&vote); err != nil {
```
acroca: Same here, should we log?

kusd/validator/validator.go
```diff
@@ -31,8 +32,30 @@ type Backend interface {
...
+	SetDeposit(deposit uint64)
+	Pending() (*types.Block, *state.StateDB)
+	PendingBlock() *types.Block
+	AddProposal(proposal *types.Proposal) error
+	AddVote(vote *types.Vote) error
+	AddBlockFragment(blockNumber *big.Int, round uint64, fragment *types.BlockFragment) error
+}
+
+var (
+	ErrCantStopNonStartedValidator       = errors.New("can't stop consensus, not started")
```
acroca: just for consistency, shouldn't this be `not validating`?

kusd/handler.go
```diff
@@ -628,48 +628,36 @@ func (pm *ProtocolManager) handleMsg(p *peer) error {
...
 
+	case msg.Code == BlockFragmentMsg:
 		// Retrieve and decode the propagated block fragment
 		var request blockFragmentData
 		if err := msg.Decode(&request); err != nil {
 			return errResp(ErrDecode, "msg %v: %v", msg, err)
 		}
-
-		// @TODO (rgerades) - fragment hash?
+		if err := pm.validator.AddBlockFragment(request.BlockNumber, request.Round, request.Data); err != nil {
```
acroca: Same here, log the error?

kusd/handler.go
jmatosp: I think so, either log or comment saying why we ignoring the error

kusd/handler.go
jmatosp: :+1:

kusd/handler.go
jmatosp: :+1:

kusd/validator/validator.go
jmatosp: :+1:  true one or the other, not sure which though

kusd/handler.go
rgeraldes: on the same file pm.txpool.AddRemotes(txs) throws an error but they are not handling the error. Have a look at their flow to see if we can do something similar.

kusd/validator/validator.go
```diff
@@ -30,750 +18,16 @@ type Backend interface {
...
+	Stop() error
+	SetExtra(extra []byte) error
+	Validating() bool
+	SetCoinbase(addr common.Address) error
+	SetDeposit(deposit uint64)
+	Pending() (*types.Block, *state.StateDB)
+	PendingBlock() *types.Block
+	AddProposal(proposal *types.Proposal) error
+	AddVote(vote *types.Vote) error
+	AddBlockFragment(blockNumber *big.Int, round uint64, fragment *types.BlockFragment) error
```
rgeraldes: LIKE!

kusd/handler.go
jmatosp: not sure if I follow, they are ignoring the errors, what should we do then?

kusd/handler.go
```diff
@@ -628,48 +628,39 @@ func (pm *ProtocolManager) handleMsg(p *peer) error {
 		pm.txpool.AddRemotes(txs)
 
 	case msg.Code == ProposalMsg:
-		// @TODO (rgeraldes) - review flow (we will not need this condition)
-		if !pm.validator.Validating() {
```
rgeraldes: The flow changed a little bit - We don't need to decode the message if we're not validating. 
With the new flow we decode the message even if it's not necessary.

jmatosp: this should be handled in a diff way, but I see why we need the Validating(), adding back

PR #116, opened 2018-02-27
Wallet/Account pair

jmatosp:
This will allow to use a valida Wallet/Account pair in validator, reducing the dependency on AccountManager()

No impact on existing functionality


accounts/wallet_account.go
```diff
@@ -0,0 +1,32 @@
...
+	Account() Account
+}
+
+type walletAccount struct {
+	Wallet
+	account Account
+}
+
+var (
+	errInvalidAccountAddress = errors.New("invalid account address, doesnt exists in wallet")
```
rgeraldes: we should move this one to the beginning of the file - they follow that convention in go-eth afaik

rgeraldes: also I think that it would be important to show the the account address that is invalid.

jmatosp: :+1: 
will make the changes

accounts/wallet_account.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (err errInvalidAccountAddress) Error() string {
+	return fmt.Sprintf("invalid account address, doesnt exists in wallet: %s" + err.account.Address.String())
+}
+
+func newErrInvalidAccountAddress(account Account) errInvalidAccountAddress {
+	return errInvalidAccountAddress{account}
+}
+
```
rgeraldes: noticed just know that the accounts package contains an errors.go file with all the errors.

jmatosp: will move there

PR #125, opened 2018-03-02
replaced a panic log with a warn on validator start if coinbase is no…

### LINKED ISSUE #124: [kusd] "failed to get wallet account"
### _Type_:
Issue

If I run an archival node, it throws this error. Previously this validation was only done on the validator start.

./kusd --config /Users/ricardogeraldes/Code/src/github.com/kowala-tech/kUSD/internal/assets/sample-kowala3.toml init /Users/ricardogeraldes/Code/src/github.com/kowala-tech/kUSD/internal/assets/genesis.json
./kusd --config /Users/ricardogeraldes/Code/src/github.com/kowala-tech/kUSD/internal/assets/sample-kowala3.toml --bootnodes "enode://a46e5885d3da52bf452fcea7269b91059d93819e7e906cb5d29575508a18a4009e5d688bf29abf8cdd55aad65a9ca4beab98de8c0a226d694c8140acdedf3a55@18.219.208.190:33448,enode://9791d40d97585d7241edeedd2491187e82e3273414cf1bed2aca049e8b21d96930437efe3e591d6996ee303d3e2fe4338e53551bb6e50f05dfeeaccd660eac01@18.220.38.32:33448" --verbosity 4

### _Affected servers or services_:
Archival node + validator node without key related flags.

### _Priority_:
High

jmatosp:
Fixes node start failure on validator is a coinbase is not set on startup

Closes #124

### DIFF
--- a/kusd/backend.go
+++ b/kusd/backend.go
@@ -145,7 +145,7 @@ func New(ctx *node.ServiceContext, config *Config) (*Kowala, error) {
 	networkContract := getNetworkContract(kusd.BlockChain(), NewContractBackend(kusd.ApiBackend))
 	walletAccount, err := getWalletAccount(ctx.AccountManager, kusd.coinbase)
 	if err != nil {
-		log.Crit("failed to get wallet account", "err", err)
+		log.Warn("failed to get wallet account", "err", err)
 	}
 	kusd.validator = validator.New(walletAccount, kusd, networkContract, kusd.chainConfig, kusd.EventMux(), kusd.engine, vmConfig)
 	kusd.validator.SetExtra(makeExtraData(config.ExtraData))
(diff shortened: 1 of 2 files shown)


rgeraldes:
I think that we should aim for something different - we're trying to do an operation that should not be done if an account was not provided. Note also that you are replacing the initial WalletAccount with another one at start - so there's not much sense on including it during the validator creation.

rgeraldes:
One more thing that we can cover: ERROR[03-02|13:01:58] Error stopping Consensus                 err="can't stop validator, not started" - We should not log this as an error.

jmatosp:
Yes, I agree. We should not start or even instantiate Validator is it is not requested to run.

So would be a good solution to lazy create a Validator on api.Start ?

Is so we could simplify the Validator process, and Validator instantiated would mean Validator running

jmatosp:
Regarding the WalletAccount being replaced, that also a workaround because of the method signature, dont think in the current form Start() needs any parameters.

When we SetCoinbase or SetDeposit it delegates this to validator, so validator always has updates values

PR #128, opened 2018-03-02
Feature/drone rename

yourheropaul:
This is a tiny fix that renames the Drone.io Slackbot to 'kUSD' (it was 'core', which is the same as s different Slackbot that we also use).


jmatosp:
Is it for master? has a lot of changes including a breaking one I think

yourheropaul:
This should be for dev, not master.

PR #129, opened 2018-03-02
Release 1.0.4

yourheropaul:
This is the 1.0.4 release candidate.

Important changes:
- [x] Both full and fast sync modes are should now be working. 

Mandatory:
- [x] No additional documentation to add.


yourheropaul:
Does this include any breaking changes?

jmatosp:
Not entirely sure, but I think that it breaks on archive nodes starting. (or any node without coinbase set) 
@rgeraldes

yourheropaul:
I'm going to remove this and merge the hotfix instead.

PR #146, opened 2018-03-15
Feature/validators network

jmatosp:
Moved network contract calls to interface


kusd/validator/states.go
```diff
@@ -71,7 +70,7 @@ func (val *validator) notLoggedInState() stateFn {
 		}
 	} else {
 		// sanity check
-		isVoter, err := val.network.IsVoter(&bind.CallOpts{}, val.walletAccount.Account().Address)
```
acroca: The `walletAccount` now seems to be stored in the `validationNetwork` right? What if that value changes in the `validator` but the network didn't get the change? in other words, are we making sure the `network` uses the same wallet account as the validator? I think calling functions that update the `walletAccount` like the `SetCoinbase` won't update the `network` object with the new account.

jmatosp: you are absolutely right, I just ignored that :)

I can quick fix, but it need bigger changes, next refactor...

kusd/validator/validation_network.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/kowala-tech/kUSD/common"
+	"github.com/kowala-tech/kUSD/contracts/network"
+	"github.com/kowala-tech/kUSD/core/types"
+	"math/big"
+)
+
+type ValidatorsChecksum [32]byte
+
+// ValidationNetwork is a gateway to validators contracts on the network
+type ValidationNetwork interface {
```
rgeraldes: By looking at this interface, I immediately thought about this one: https://github.com/kowala-tech/kUSD/blob/master/consensus/consensus.go#L35
There's a large part of the validator state (related to consensus) that eventually could be ported to our implementation of the tendermint consensus engine. Something like proof of activity protocol - https://github.com/ethereum/go-ethereum/tree/master/consensus/clique
From a testing point of view I'm not sure if it's the best solution.

jmatosp: Not sure if I follow :|
But if we can port, and it's stable then we should.

This change actually allows to test the Validator without needing access to network contracts, thru it's interface

rgeraldes: It's something can be done later on (if necessary) - I will open an issue.

PR #161, opened 2018-03-20
consensus: unbonding period and contract tests

### LINKED ISSUE #100: [console] add minimum deposit methods
### _Type_:
Request

### _What do you need?_:
- getMinDeposit - (depends on sync) returns the current minimum deposit if 
there are positions available / or the smallest deposit if all the positions have been filled.
- setDeposit - sets the deposit

If any of the network values change before submitting the transaction, we should abort and let the user know - the user avoids spending gas for a possible unsuccessful operation.

### _Priority_:
High

### LINKED ISSUE #141: [console] complete getMinimumDeposit method (validator module)
### _Type_:
Request

### _What do you need?_:
Get the value based on the contract instead of the config.

### _Priority_:
Medium

### _Other info we may need_:
Depends on #134

rgeraldes:
PR Closes #136, Closes #100, Closes #141.

**Important changes:**
- Included the latest version of the contract - includes deposits + unbonding period.
- Removed the miner api (obsolete - no longer used) from kusd/api.go.
- Moved validator/validationNetwork (wrapper type) and the interface closer to the contract (go-eth style) and renamed it to "election".
- removed availability check (join method) - https://github.com/kowala-tech/kUSD/blob/dev/kusd/validator/validation_network.go#L39 - it's no longer necessary since everyone can join the election as long as the user deposits the minimum required deposit.
- Added two methods to the validator api: getDeposits & redeemDeposits.
- Added to the genesis creation(puppeth): max validators, genesis address, unbonding period, base deposit.
- removed contracts that are not being used/not finalised. (contracts can be found here: 
https://github.com/kowala-tech/contracts)
- added contracts/network to the stable packages for tests. 
- modified GetMinimumDeposit to use the contract vs config.
- included the latest genesis file in internal/assets.

**Note:**

- It's not going to be possible to use the default testnet genesis for now since we include accounts with storage and code and the ethereum tool cannot handle that for now - https://github.com/kowala-tech/kUSD/blob/dev/core/mkalloc.go#L53. We will need to perform the "kusd init" step.

**Post:** 

- Create issue for this idea (core/types/validator.go)
Deposit would sit nice under Validator. Seems like a good option for an Aggregate with Validator as top level - https://martinfowler.com/bliki/DDD_Aggregate.html
- Add logic to redeemDeposits (contracts/network/election.go) to skip the operation if the user does not have at least one unlocked deposit.


contracts/network/contracts/Election.sol
```diff
@@ -0,0 +1,224 @@
...
+        validatorsChecksum = keccak256(validatorIndex);
+    }
+
+    function _insertValidator(address code, uint deposit) private {
+        Validator sender = validators[code];
+        sender.index = validatorIndex.push(code) - 1;
+        sender.isValidator = true;
+        sender.deposits.push(Deposit({amount:deposit, releasedAt: 0}));
+
+        for (uint index = sender.index; index > 0; index--) {
```
acroca: this seems to be looping for all the rest of validators and pushing the new one down in the list keeping it sorted by their latest collateral amount, right? wouldn't it be clearer to loop from `sender.index - 1` to `index == 0`? Also, just for readability, we could move this to a separate function or add a comment to this loop.

contracts/network/election.go
```diff
@@ -0,0 +1,160 @@
+package network
+
+//go:generate solc --abi --bin --overwrite -o build contracts/Election.sol
+//go:generate abigen -abi build/Election.abi -bin build/Election.bin -pkg contracts -type ElectionContract -out contracts/election.go
```
acroca: Shouldn't the output be `contracts/network/election.go`?

core/types/validator.go
```diff
@@ -87,3 +87,15 @@ func (set *ValidatorSet) Contains(addr common.Address) bool {
...
+func NewDeposit(amount *big.Int) *Deposit {
+	return &Deposit{
+		amount: amount,
+	}
+}
+
+// Deposit represents the validator deposits at stake
+type Deposit struct {
+	amount *big.Int
+	//releasedAt
```
acroca: Any reason why `amount` is private? this type is returned by the `GetDeposits`, if we don't expose it somehow it's not very useful to get the deposits. Also, any reason why the `ReleasedAt` is not included?

kusd/api.go
```diff
@@ -170,6 +110,18 @@ func (api *PrivateValidatorAPI) GetMinimumDeposit() uint64 {
 	return api.kusd.GetMinimumDeposit()
 }
 
+// GetDeposits returns the validator deposits
+func (api *PrivateValidatorAPI) GetDeposits() ([]*types.Deposit, error) {
+	return api.kusd.Validator().Deposits()
+}
+
+func (api *PrivateValidatorAPI) RedeemDeposits() (bool, error) {
```
acroca: Probably this could just return `error`.

kusd/validator/validator.go
```diff
@@ -71,7 +73,7 @@ type validator struct {
 
 	walletAccount accounts.WalletAccount
 
-	network ValidationNetwork
+	election network.Election // consensus election
```
acroca: I find it confusing to have `validator.election` using `network.Election` and also `Election` from the current package a few lines above. Not sure what is what. Could we use a more specific naming to differentiate them? Maybe for a future refactor PR.

contracts/network/contracts/Election.sol
```diff
@@ -0,0 +1,224 @@
...
+        uint releasedAt;
+    }
+
+    // Validator represents a consensus validator      
+    struct Validator {
+        uint index;
+        bool isValidator;
+
+        // @NOTE (rgeraldes) - users can have more than one deposit
+        // Example: user leaves and re-enters the election. At this point
```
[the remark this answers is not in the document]
rgeraldes: Depends: The coins are locked for a certain time, so it's possible that if he re-enters right away he can end up with two locked deposits but the first one with a release date. The unboding period (time locked) will be set during the genesis creation.

consensus/tendermint/reward.go
```diff
@@ -1,14 +1,6 @@
...
-import (
-	"fmt"
-	"math/big"
-
-	"github.com/kowala-tech/kUSD/common"
-	"github.com/kowala-tech/kUSD/contracts/network"
-	"github.com/kowala-tech/kUSD/core/state"
-	"github.com/kowala-tech/kUSD/params"
-)
+/*
```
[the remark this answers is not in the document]
rgeraldes: I though about it - I will do it.

contracts/network/contracts/Election.sol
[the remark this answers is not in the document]
acroca: I got that, but if he re-enters the election we could release the funds right away. What's the reason to keep two locked funds? leaving and re-entering doesn't seem to be a way of attacking the network, as long as the deposit is paid (once) it's safe, no?

contracts/network/election.go
rgeraldes: I followed ethereum's way -This process generates the contract bindings for go. The bindings usually stay in the same package as the contract and the wrapper type stays in the network package.

contracts/network/contracts/Election.sol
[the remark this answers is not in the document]
rgeraldes: Tendermint is aiming for a unbonding period of 4 weeks.

contracts/network/election.go
```diff
@@ -0,0 +1,171 @@
...
+	IsGenesisValidator(address common.Address) (bool, error)
+	IsValidator(address common.Address) (bool, error)
+}
+
+type election struct {
+	*contracts.ElectionContract
+	chainID *big.Int
+}
+
+func getContractAddr(chainID *big.Int) common.Address {
```
jmatosp: Does a map of global static ChainID -> contractAddress removes this function?

contracts/network/election.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func NewElection(contractBackend bind.ContractBackend, chainID *big.Int) (*election, error) {
+	contract, err := contracts.NewElectionContract(getContractAddr(chainID), contractBackend)
+	if err != nil {
+		return nil, err
+	}
+
+	return &election{
+		ElectionContract: contract,
```
jmatosp: why exported? and not just contract, it's already named `election` object

contracts/network/election.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		weight := big.NewInt(0)
+		validators[i] = types.NewValidator(validator.Code, validator.Deposit.Uint64(), weight)
+	}
+
+	return types.NewValidatorSet(validators), nil
+}
+
+func (election *election) Deposits(addr common.Address) ([]*types.Deposit, error) {
```
jmatosp: this could be removed, if we moved the `Deposit` object to the `Validator`,  they should be bounded together.

core/types/validator.go
```diff
@@ -87,3 +88,24 @@ func (set *ValidatorSet) Contains(addr common.Address) bool {
 	_, ok := set.membership[addr]
 	return ok
 }
+
+func NewDeposit(amount *big.Int, unixTimestamp int64) *Deposit {
```
jmatosp: this would sit nice under Validator. Seems like a good option for an Aggregate with Validator as top level

https://martinfowler.com/bliki/DDD_Aggregate.html

kusd/api.go
jmatosp: :+1:

kusd/backend.go
```diff
@@ -142,12 +142,15 @@ func New(ctx *node.ServiceContext, config *Config) (*Kowala, error) {
 	kusd.ApiBackend.gpo = gasprice.NewOracle(kusd.ApiBackend, gpoParams)
 
 	// consensus validator
-	networkContract := getNetworkContract(kusd.BlockChain(), NewContractBackend(kusd.ApiBackend))
+	election, err := network.NewElection(NewContractBackend(kusd.ApiBackend), chainConfig.ChainID)
+	if err != nil {
+		log.Crit("Failed to load the network contract", "err", err)
```
jmatosp: do we Panic, or should we just not start the Validator?

kusd/validator/validator.go
```diff
@@ -49,11 +49,13 @@ type Validator interface {
 	AddProposal(proposal *types.Proposal) error
 	AddVote(vote *types.Vote) error
 	AddBlockFragment(blockNumber *big.Int, round uint64, fragment *types.BlockFragment) error
+	Deposits() ([]*types.Deposit, error)
```
jmatosp: we are prob going to need to split this interface, just thinking how and where

kusd/validator/validator.go
jmatosp: still don't understand whats what :)

maybe we should not call Election the network one, and just network.contract,  for me moment or until better name comes

the embedded Election it's just data, no behaviour :|, also need cleanup most is not in use. `VotingState`?

kusd/api.go
rgeraldes: They use this pattern all over the API - example: https://github.com/ethereum/go-ethereum/blob/master/eth/api.go#L214. I think that in the end the goal is provide some input to the user - It's probably not the best way to it though. What do you think about it? @acroca @jmatosp

kusd/validator/validator.go
rgeraldes: yes, the contract one is the real election and the other one as you say is election internal state.
I'd probably change the internal state name - VotingState sounds good.

contracts/network/contracts/Election.sol
```diff
@@ -0,0 +1,227 @@
+pragma solidity 0.4.21;
```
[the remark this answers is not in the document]
rgeraldes: yes, I've added all the info in here - generate bindings, warnings and common pitfalls - https://github.com/kowala-tech/kUSD/blob/a2bd9b4ced68330434a3c63731dbd67d63b7ec75/contracts/doc.go

contracts/network/contracts/Election.sol
```diff
@@ -0,0 +1,227 @@
...
+    uint public unbondingPeriod;
+    address public genesisValidator;
+
+    // validatorsChecksum is a representation of the current set of validators
+    bytes32 public validatorsChecksum;
+
+    // Deposit represents the collateral - staked tokens
+    struct Deposit {
+        uint amount;
+        uint releasedAt;
```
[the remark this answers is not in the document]
rgeraldes: availableAt sounds good. I will make the change.

kusd/api.go
acroca: I prefer just to return an error, the bool is really just `err != nil`

kusd/backend.go
rgeraldes: We need to panic as the validator & api depend on this interface - even if the validator is not running you may need to use this interface (ex: redeemDeposits()).

contracts/network/contracts/Election.sol
```diff
@@ -0,0 +1,227 @@
...
+        deposit = validator.deposits[validator.deposits.length - 1].amount;
+    }
+
+    function _hasAvailability() public view returns (bool available) {
+        return (maxValidators - validatorIndex.length) > 0;
+    }
+
+    // getMinimumDeposit returns the base deposit if there are positions available or
+    // the current smallest deposit required if there aren't positions availabe.
+    function getMinimumDeposit() public view returns (uint deposit) {
```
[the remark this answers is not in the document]
rgeraldes: What would you change? It's not clear to me.

kusd/validator/validator.go
rgeraldes: :)

rgeraldes:
One more: Added contracts/network to the stable packages for tests.

rgeraldes:
I've modified GetMinimumDeposit to use the contract vs config.

rgeraldes:
Documentation was updated

kusd/api.go
```diff
@@ -159,17 +99,56 @@ func (api *PrivateValidatorAPI) SetCoinbase(coinbase common.Address) bool {
...
+func (api *PrivateValidatorAPI) GetMinimumDeposit() (uint64, error) {
 	return api.kusd.GetMinimumDeposit()
 }
 
+// GetDepositsResult is the result of a validator_getDeposits API call.
+type GetDepositsResult struct {
+	Deposits []depositEntry `json:"deposits"`
+}
+
+type depositEntry struct {
```
acroca: `GetDepositsResult` is an exposed struct with an array of `depositEntry`. Shouldn't this also be exposed for consistency?

core/types/validator.go
```diff
@@ -100,3 +102,24 @@ func (set *validatorList) Contains(addr common.Address) bool {
 	validator := set.Get(addr)
 	return validator != nil
 }
+
+func NewDeposit(amount uint64, unixTimestamp int64) *Deposit {
```
acroca: Aren't not all amounts `big` numbers?

rgeraldes: The motivation for this one is that for the console cannot display *big.Int. The other solution is to force a deposit to be a multiplier of the ether denomination and the deposit would be showed in ether. Overflow is possible in the current option since the ether multiplier is 1*10^18. Maybe we should go with the other option and this code would stay like this as 1^10*18 ether is not possible - coins in circulation.

kusd/api.go
rgeraldes: followed the convention in the same file: https://github.com/kowala-tech/kUSD/blob/0c3a346694015e3bea93ec408f2e8ee1365b3ed6/kusd/api.go#L540

core/types/validator.go
rgeraldes: I've modified the deposits to return the values in ether (default is wei). In the future we can modify setDeposits to take this into account and also the cmdline flag otherwise this issue will be infinite :)

contracts/network/contracts/Election.sol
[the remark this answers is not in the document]
jmatosp: can you write one?

kusd/backend.go
jmatosp: That give a new perspective, we can access the Validation network even if we are not a validator or validating. So in future I think we going to end up with 2 diff interfaces

contracts/network/contracts/Election.sol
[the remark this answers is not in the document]
jmatosp: ignore me

cmd/puppeth/wizard_genesis.go
```diff
@@ -77,78 +65,68 @@ func createContract(cfg *runtime.Config, code []byte) (*contractData, error) {
...
+			Tracer: newVmTracer(),
+		},
 	}
-	netMapParams, err := netMapAbi.Pack("", r[1].addr, r[2].addr, r[0].addr)
+
+	contract, err := createContract(runtimeCfg, append(common.FromHex(contracts.ElectionContractBin), electionParams...))
 	if err != nil {
-		fmt.Println("can't pack network map contract params:", err)
-		os.Exit(-11)
+		log.Error("can't create the election contract", err)
```
jmatosp: we should either log or return not both

cmd/puppeth/wizard_genesis.go
```diff
@@ -77,78 +65,68 @@ func createContract(cfg *runtime.Config, code []byte) (*contractData, error) {
...
+
+	electionParams, err := electionABI.Pack("", baseDeposit, maxVal, unbondingPeriod, *genesisAddr)
 	if err != nil {
-		fmt.Println("can't pack price oracle contract params:", err)
-		os.Exit(-8)
-	}
-	if c, err = createContract(runtimeConfig, append(contractsCode[ctPriceOracle], priceOracleParams...)); err != nil {
-		fmt.Println("can't create price oracle contract:", err)
-		os.Exit(-9)
+		log.Error("can't pack the election contract params", err)
```
jmatosp: log or return err

cmd/puppeth/wizard_genesis.go
```diff
@@ -77,78 +65,68 @@ func createContract(cfg *runtime.Config, code []byte) (*contractData, error) {
...
 	}
-	r = append(r, c)
-	// create price oracle contract
-	priceOracleAbi, err := abi.JSON(strings.NewReader(network.PriceOracleContractABI))
+
+	electionABI, err := abi.JSON(strings.NewReader(contracts.ElectionContractABI))
 	if err != nil {
-		fmt.Println("can't parse price oracle contract ABI:", err)
-		os.Exit(-7)
+		log.Error("can't parse the election contract ABI:", err)
```
jmatosp: log or return err

PR #163, opened 2018-03-21
renamed validators to voters

jmatosp:
New naming for voters and reducing interface of voters


rgeraldes:
I think that this change does not make sense from a functional point of view as the accepted term for the voting entity is Validator - not sure if there's a technical reason behind it.

jmatosp:
You are right, they are voters, I'm experimenting in trying to have different naming for Validator as there are lots. This object is used in a VotingTable and VotingSystem so I thought that in that context they are voters.

But happy to close

jmatosp:
The main change here is actually NextProposer :) suggested by @acroca and doest allow us to high the UpdateWeight implementation

Issue #68, opened 2018-02-08
End-to-end tests


yourheropaul:
This will require some scenarios and the containerised testnet.

PR #164, opened 2018-03-22
Feature/e2e

acroca:
Adds e2e tests to the project.

I moved the Vendor commit to be the first one so it can be easily ignored in the github review interface.

Important changes:
- [x] Added clustering tools to create and interact with a kubernetes cluster.
- [x] Added first simple feature test using [godog](https://github.com/DATA-DOG/godog)


jmatosp:
Can you add the new packages to `CI.go` stable?

.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
```
jmatosp: don't think this file doesn't belong to the repo

rgeraldes:
#68

rgeraldes:
I will review it later today/tomorrow

Makefile
```diff
@@ -175,3 +175,13 @@ docker-publish-kusd:
 
 docker-publish-faucet:
 	docker push kowalatech/faucet
+
+## E2E tests
+
+GODOG_BIN := $(shell command -v godog 2> /dev/null)
+e2e:
```
rgeraldes: do we need to install godog everytime if we just want to execute tests?

cluster/backend.go
```diff
@@ -0,0 +1,28 @@
...
+	log.Println("Connecting to the k8s cluster")
+	config, err := cluster.RestConfig()
+	if err != nil {
+		return nil, err
+	}
+	return kubernetes.NewForConfig(config)
+}
+
+func (cluster *k8sCluster) RestConfig() (*rest.Config, error) {
+	kubeconfig := filepath.Join(os.Getenv("HOME"), ".kube", "config")
```
rgeraldes: There are some constants on this line

cluster/bootnode.go
```diff
@@ -0,0 +1,114 @@
...
+		return "", fmt.Errorf("Bootnode enode not found in the logs.")
+	}
+
+	enodeSecret := string(found[1])
+	ip := pod.Status.PodIP
+
+	return fmt.Sprintf(`enode://%v@%v:%v`, enodeSecret, ip, bootnodePort), nil
+}
+
+func randStringBytes(n int) string {
```
rgeraldes: move this one to utils.go & remove extra lines?

cluster/cluster.go
```diff
@@ -0,0 +1,153 @@
...
+
+func (client *cluster) waitForKusdPod(podName string) error {
+	log.Printf("Waiting for pod `%v`...\n", podName)
+	return WaitFor(2*time.Second, 2*time.Minute, func() bool {
+		return client.isKusdPodRunning(podName)
+	})
+}
+
+func (client *cluster) GetBalance(podName string) (*big.Int, error) {
+	resp, err := client.Exec(podName, fmt.Sprintf(`eth.getBalance(eth.coinbase)`))
```
rgeraldes: is this one using the console?

rgeraldes:
Overall I think that we are relying on the console for some steps and we probably should rely on the APIs instead as the console usage is under discussion and it might change.

Makefile
acroca: The godog installation is triggered only if the binary is not found.

acroca:
@rgeraldes I think it's better to use the console in e2e tests, since it's closer to what the user would do.

cluster/cluster.go
acroca: yes

jmatosp:
Both have a good point, but I would prefer API directly as it will probably better to refactor (if using smart IDE).

But either way this should go forward so we can have some sort of tests.

PR #165, opened 2018-03-22
refactoring to smaller voting table

jmatosp:
Cleanup Voting Table
Tests for Voting Table behaviour


core/voting_table.go
```diff
@@ -45,208 +41,72 @@ import (
...
 		}
+	}
+	return false
+}
 
+func (table *VotingTable) isVoterParticipating(address common.Address) bool {
+	return table.voters.Contains(address)
+}
+
+func (table *VotingTable) haveQuorum() bool {
```
rgeraldes: haveQuorum > hasQuorum?

core/voting_table.go
```diff
@@ -45,208 +41,72 @@ import (
...
 	from, err := types.VoteSender(table.signer, vote) // already validated & cached
 	if err != nil {
-		return false, err
+		return err
 	}
 
-	index, ok := table.addressToIndex[from]
-	if !ok {
-		// @TODO (rgeraldes)
+	if !table.isVoterParticipating(from) {
```
rgeraldes: isVoterParticipating  - IsVoter (should be enough)

core/voting_table.go
```diff
@@ -45,208 +41,72 @@ import (
...
+		return errors.New("voter address not found in voting table")
```
rgeraldes: what do you think about adding the address to the log?

core/types/vote.go
```diff
@@ -28,6 +28,36 @@ func (typ VoteType) IsValid() bool {
 	return typ >= PreVote && typ <= PreCommit
 }
 
+type SignedVote interface {
```
jmatosp: temp till find a better way

kusd/validator/election.go
```diff
@@ -76,19 +79,29 @@ func NewVotingSystem(eventMux *event.TypeMux, signer types.Signer, electionNumbe
...
 
 // Add registers a vote
-func (vs *VotingSystem) Add(vote *types.Vote, local bool) (bool, error) {
-	// @TODO (rgeraldes) - validation
+func (vs *VotingSystem) Add(vote *types.Vote) error {
 	votingTable := vs.getVoteSet(vote.Round(), vote.Type())
-	votingTable.Add(vote, local)
 
-	return false, nil
+	signedVote, err := types.NewSignedVote(vs.signer, vote)
```
rgeraldes: Instead of NewSignedVote I think that vote.AsSignedVote makes more sense - but if the purpose is to get the sender I think that we should look for a different solution. It seems that we are creating a new vote.

core/voting_table.go
```diff
@@ -1,252 +1,72 @@
...
-		log.Trace("Discarding invalid transaction", "hash", hash, "err", err)
-		invalidTxCounter.Inc(1)
-		return false, err
-	}
-
-	/*
-		// Check signature.
-		if err := vote.Verify(voteSet.chainID, val.PubKey); err != nil {
-			return false, errors.Wrapf(err, "Failed to verify vote with ChainID %s and PubKey %s", voteSet.chainID, val.PubKey)
+func (table *votingTable) seen(vote *types.Vote) bool {
```
rgeraldes: seen vs duplicate

rgeraldes: from an election point of view - I hear the word "duplicate" a lot.

kusd/validator/election.go
jmatosp: yeah, I commented myself on this one :)

it will prob go away as I refactor outer layers.

Yes, `Vote` should prob have `Address`? 

if I have an instance of `Vote` then it's a valid one (signed) so the consumer of vote doesn't need to worry about checking is valid.

core/voting_table.go
rgeraldes: sounds good!

kusd/validator/election.go
jmatosp: Can we add Address to Vote and on decoding a vote fail if signature is wrong

rgeraldes: transaction does something like that from field - https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go#L54

jmatosp: will something similiar on next refactor

core/voting_table.go
```diff
@@ -1,252 +1,72 @@
...
-		eventMux:      eventMux,
-		signer:        signer,
-		// @TODO (rgeraldes) - replace with constants
-		quorum:         voters.Size()*2/3 + 1,
-		addressToIndex: make(map[common.Address]int, voters.Size()),
-		//maj23:         nil,
-		//peerMaj23s:    make(map[string]BlockID),
+func (table *votingTable) Add(vote types.SignedVote) error {
+	if !table.isVoter(vote.Address()) {
+		return errors.New("voter address not found in voting table")
```
rgeraldes: include address in the error?

core/voting_table.go
```diff
@@ -1,252 +1,72 @@
...
 	}
 
-	// cache voter index
-	for i := 0; i < table.voters.Size(); i++ {
-		table.addressToIndex[table.voters.At(i).Address()] = i
+	if table.isDuplicate(vote.Vote()) {
+		return errors.New("conflict vote seen before")
```
rgeraldes: errDuplicateVote = errors.new("duplicate vote")

core/voting_table_test.go
```diff
@@ -1,3 +1,117 @@
...
+		{3, 3, true},
+		{10, 6, false},
+		{10, 7, true},
+		{100, 1, false},
+		{100, 66, false},
+		{100, 67, true},
+		{100, 100, true},
+	}
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("voters %d votes %d quorum %t", tc.voters, tc.votes, tc.hasQuorum), func(t *testing.T) {
```
rgeraldes: I like this approach to naming the test - I usually add a name/context field but this makes more sense in some situations.

PR #169, opened 2018-03-26
New network ID

yourheropaul:
This changes the ChainID / NetworkID for testnet mining clients. I've incremented the number by 1.


rgeraldes:
We also need to change this value: https://github.com/kowala-tech/kUSD/blob/dev/params/config.go#L25

yourheropaul:
Okay, also updated `params/config.go`. Anything else missing?

PR #172, opened 2018-03-26
[WIP] bug/replaced isValidator with TX status (tx failure)

### LINKED ISSUE #160: [consensus] replace isValidator with tx status
### _Type_:
Request

### _If it's a request, what do you need it for?_:
The program hangs forever if the transaction status is not ok (register validator).
The validator never leaves if the transaction status is not ok (deregister validator).
The user never gets his deposit back if the transaction status is not ok (redeem deposits).

### _Priority_:
Medium

rgeraldes:
PR Closes #160 

- Replaced isValidator with TX status for the transactional methods - join, withdraw. The previous version checks the contract for confirmation, but in the end the confirmation is done one the transaction level and not on the contract. Transaction could fail and we would loop forever the contract to get some sort of positive confirmation - the negative case (transaction failed) is not covered by the contract.
- IsValidator is still going to be used by the state machine to know when to leave.

### DIFF
--- a/kusd/validator/states.go
+++ b/kusd/validator/states.go
@@ -38,7 +38,8 @@ func (val *validator) notLoggedInState() stateFn {
 		chainHeadSub := val.chain.SubscribeChainHeadEvent(chainHeadCh)
 		defer chainHeadSub.Unsubscribe()
 
-		if err := val.network.Join(val.walletAccount, val.deposit); err != nil {
+		tx, err := val.network.Join(val.walletAccount, val.deposit)
+		if err != nil {
 			log.Error("Error joining validators network", "err", err)
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


kusd/validator/states.go
```diff
@@ -52,14 +53,20 @@ func (val *validator) notLoggedInState() stateFn {
 					return nil
 				}
 
-				confirmed, err := val.network.IsVoter(val.walletAccount.Account().Address)
-				if err != nil {
-					log.Crit("Failed to verify the voter registration", "err", err)
+				// checks whether a transaction has been included or not
```
jmatosp: why not make this as part of the `network.Join`?

kusd/validator/validator.go
```diff
@@ -378,11 +380,41 @@ func (val *validator) commitTransaction(tx *types.Transaction, bc *core.BlockCha
...
 
-func (val *validator) withdraw() {
-	err := val.network.Withdraw(val.walletAccount)
+func (val *validator) withdraw() error {
+	tx, err := val.network.Withdraw(val.walletAccount)
 	if err != nil {
 		log.Error("failed to withdraw from the election", "err", err)
 	}
+
+	// wait until transaction is successful/failed
```
jmatosp: same here, why not make part of the `network.Withdraw`?

PR #177, opened 2018-03-27
New genesis

yourheropaul:
New genesis.json for updated testnet. This was generated with a new build of puppeth.

Is there anything missing or that shouldn't be here?


acroca:
I think this new genesis has only 1 contract instead of 4. Is that correct? @rgeraldes

rgeraldes:
confirmed @acroca - we have been removing unfinished solutions (previous contracts) from the codebase.

PR #209, opened 2018-04-04
refactored validators to voters

jmatosp:
Renaming Validators to votes
Test deposit and Hash of Voters


core/types/voters.go
```diff
@@ -0,0 +1,152 @@
...
+import (
+	"errors"
+	"math/big"
+
+	"github.com/kowala-tech/kcoin/common"
+	"github.com/kowala-tech/kcoin/rlp"
+	"io"
+)
+
+// Validator represents a consensus Voter
```
rgeraldes: Validator > Voter

core/types/voters.go
```diff
@@ -0,0 +1,152 @@
...
+type Voters interface {
+	NextProposer() *Voter
+	At(i int) *Voter
+	Get(addr common.Address) *Voter
+	Len() int
+	Contains(addr common.Address) bool
+	Hash() common.Hash
+}
+
+var ErrInvalidParams = errors.New("voters set needs at least one voter")
```
rgeraldes: should be moved to the top of the file?

core/types/voters.go
```diff
@@ -0,0 +1,152 @@
...
+
+	set := &voters{
+		voters: voterList,
+	}
+
+	return set, nil
+}
+
+// voters is a list of Voter
+type voters struct {
```
rgeraldes: type voters []*Vote?

core/types/voters_test.go
```diff
@@ -0,0 +1,147 @@
...
+	assert.Equal(t, voter, voters.NextProposer())
+}
+
+func TestVoters_UpdateWeightChangesProposer(t *testing.T) {
+	voter1 := makeVoter("0x5aaeb6053f3e94c9b9a09f33669435e7ef1beaed", 100, 100)
+	voter2 := makeVoter("0x6aaeb6053f3e94c9b9a09f33669435e7ef1beaed", 101, 101)
+	voter3 := makeVoter("0x7aaeb6053f3e94c9b9a09f33669435e7ef1beaed", 99, 99)
+
+	voters, err := NewVoters([]*Voter{voter1, voter2, voter3})
+
```
rgeraldes: extra line

core/types/voters_test.go
```diff
@@ -0,0 +1,147 @@
...
+func TestVoters_EmptyReturnsError(t *testing.T) {
+	voters, err := NewVoters(nil)
+
+	require.Error(t, err)
+	require.Nil(t, voters)
+}
+
+func TestVoters_GetAtNegativeIndexReturnsNil(t *testing.T) {
+	voter := makeVoter("0x5aaeb6053f3e94c9b9a09f33669435e7ef1beaed", 100, 100)
```
rgeraldes: you could create a testify suite and create a voter by default

jmatosp: don't think here is the case, test suite is to for repetitive assertions? here this are normal objects not under test. 
Also the properties of Voter should be visible in the test scenario, like weight and deposit.

I did a global var with a list of voters, but don't think it reads better.

Your toughts?

PR #212, opened 2018-04-05
Feature/issue 197 non interactive genesis generator

jgimeno:
This command will be responsable to generate genesis.json files in a non-interactive way to integrate in the CI.

Important changes:

Mandatory:
- [x] Added documentation.

Closes #


jgimeno:
Don't be too bad on my first review!! ;PP just joking. 

I want to discuss some structure ideas about the genesis logic, I thought about setting the command and the handler in its own package, I was thinking about putting it in a folder called genesis/ in the root of the project, so it can look something like genesis.CreateGenesisCommand and it will be opened for everyone if they need to create it's own.

cmd/genesis/values.go
```diff
@@ -0,0 +1,48 @@
+package main
+
+import (
+	"errors"
+)
+
+const (
+	DefaultSmartContractsOwner = "0x259be75d96876f2ada3d202722523e9cd4dd917d"
```
yourheropaul: This value is derived from the [network code](https://github.com/kowala-tech/kcoin/blob/dev/contracts/network/election.go#L20), and I suggest that it only be defined only in election.go. 

In order to do that, `mapChainIDToAddr` would have to be exported and this const would have to refer to it (like `network.MapChainIDToAddr[params.TestnetChainConfig.ChainID.Uint64()]`, which is slightly ugly) but then at least the defaults will be always be up to date with the core code.

cmd/genesis/command.go
```diff
@@ -0,0 +1,304 @@
...
+	ErrEmptyUnbondingPeriod                         = errors.New("unbonding period in days is mandatory")
+	ErrEmptyWalletAddressValidator                  = errors.New("Wallet address of genesis validator is mandatory")
+	ErrInvalidWalletAddressValidator                = errors.New("Wallet address of genesis validator is invalid")
+	ErrEmptyPrefundedAccounts                       = errors.New("empty prefunded accounts, at least the validator wallet address should be included")
+	ErrWalletAddressValidatorNotInPrefundedAccounts = errors.New("prefunded accounts should include genesis validator account")
+	ErrInvalidAddressInPrefundedAccounts            = errors.New("address in prefunded accounts is invalid")
+	ErrInvalidContractsOwnerAddress                 = errors.New("address used for smart contracts is invalid")
+)
+
+type GenerateGenesisCommand struct {
```
yourheropaul: Why make this type exportable with only non-exported members? Maybe I'm misunderstanding the use cases for it, but shouldn't it be either an interface or have at least one exported member?

cmd/genesis/command.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	// Add a batch of precompile balances to avoid them getting deleted
+	for i := int64(0); i < 256; i++ {
+		genesis.Alloc[common.BigToAddress(big.NewInt(i))] = core.GenesisAccount{Balance: big.NewInt(1)}
+	}
+
+	out, _ := json.MarshalIndent(genesis, "", "  ")
+
+	_, err = h.w.Write(out)
```
yourheropaul: This comment is quite anal, but can the conditional not be one line here?

cmd/genesis/vmtracer.go
```diff
@@ -0,0 +1,38 @@
...
+	data map[common.Address]map[common.Hash]common.Hash
+}
+
+func newVmTracer() *vmTracer {
+	return &vmTracer{
+		data: make(map[common.Address]map[common.Hash]common.Hash, 1024),
+	}
+}
+
+func (vmt *vmTracer) CaptureState(env *vm.EVM, pc uint64, op vm.OpCode, gas, cost uint64, memory *vm.Memory, stack *vm.Stack, contract *vm.Contract, depth int, err error) error {
```
yourheropaul: I know this is copied from puppeth, but without tests this function is very confusing. What does it do?

jgimeno: I think this is a tracer to capture some kind of information from the real blockchain, is there anybody that can assert this info?

cmd/genesis/values.go
jgimeno: Great! Will do this.

jgimeno: I understand that this should be defined based on the network selection, main or test

kcoin/genesis/genesis_generator.go
```diff
@@ -0,0 +1,404 @@
...
+
+func setPrefundedAccounts(validPrefundedAccounts []*validPrefundedAccount, genesis *core.Genesis) {
+	for _, vAccount := range validPrefundedAccounts {
+		genesis.Alloc[*vAccount.walletAddress] = core.GenesisAccount{
+			Balance: vAccount.balance,
+		}
+	}
+}
+
+func createNetwork(network string) (string, error) {
```
jmatosp: `get`, `map` or similar on these func names, I think `create` implies creating a object or a resource, this just maps between 2 values

kcoin/genesis/genesis_generator.go
```diff
@@ -0,0 +1,404 @@
...
+	if !availableNetworks[network] {
+		return "", ErrInvalidNetwork
+	}
+
+	return network, nil
+}
+
+func createConsensusEngine(consensus string) (string, error) {
```
jmatosp: same as `createNetwork`

kcoin/genesis/genesis_generator.go
```diff
@@ -0,0 +1,404 @@
...
+
+func createUnbondingPeriod(uP string) (*big.Int, error) {
+	var text string
+	if text = strings.TrimSpace(uP); text == "" {
+		return nil, ErrEmptyUnbondingPeriod
+	}
+
+	unbondingPeriod, ok := new(big.Int).SetString(text, 0)
+	if !ok {
+		//TODO: Create error
```
jmatosp: create the error then :)

kcoin/genesis/genesis_generator.go
```diff
@@ -0,0 +1,404 @@
...
+		return nil, errors.New("invalid max num of validators.")
+	}
+
+	return unbondingPeriod, nil
+}
+
+func createWalletAddress(wA string) (*common.Address, error) {
```
jmatosp: `wA` A uppercase on purpose?

cmd/genesis/genesis.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/kowala-tech/kcoin/kcoin/genesis"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"os"
+	"strconv"
+	"strings"
+)
+
+var (
+	FileConfig string
```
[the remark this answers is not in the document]
jgimeno: right, this was the first I did and I left it. Good eye!

kcoin/genesis/genesis_generator.go
jgimeno: Thanks, I had some issues with the naming on this functions.

kcoin/genesis/genesis_generator.go
```diff
@@ -0,0 +1,404 @@
...
+	}
+
+	genesis := &core.Genesis{
+		Timestamp: uint64(time.Now().Unix()),
+		GasLimit:  4700000,
+		Alloc:     make(core.GenesisAlloc),
+		Config:    &params.ChainConfig{},
+	}
+
+	setNetwork(validOptions.network, genesis)
```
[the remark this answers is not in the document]
jgimeno: good one!

kcoin/genesis/genesis_generator.go
```diff
@@ -0,0 +1,404 @@
...
+	case MainNetwork:
+		genesis.Config.ChainID = params.MainnetChainConfig.ChainID
+	case TestNetwork:
+		genesis.Config.ChainID = params.TestnetChainConfig.ChainID
+	case OtherNetwork:
+		genesis.Config.ChainID = new(big.Int).SetUint64(uint64(rand.Intn(65536)))
+	}
+}
+
+func validateOptions(options GenesisOptions) (*validGenesisOptions, error) {
```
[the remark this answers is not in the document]
jgimeno: umm, I have my doubts in this one, since finally for me it validates if the options are correct, and if they are they return a validOptions struct. Can it be?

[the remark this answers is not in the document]
jmatosp: :) nw

does both things, depends on what you are seeing as main, does validate and map to a diff struct

You can leave as is

cmd/genesis/genesis.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func main() {
+	if err := cmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func parsePrefundedAccounts(accounts interface{}) []genesis.PrefundedAccount {
```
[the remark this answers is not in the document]
jgimeno: This is needed since data can come from 2 types, string or slice depending on where it is loaded (config or command line). I preferred to have this encapsulated in this function instead of having an if on the callee function. It is not the best, but not really really the worst. If you still want to fix this let me know! :D

kcoin/genesis/genesis_generator.go
```diff
@@ -65,7 +65,7 @@ type GenesisOptions struct {
 
 type PrefundedAccount struct {
 	WalletAddress string
-	Balance       int64
+	Balance       string
```
jmatosp: why moving to string?

cmd/genesis/genesis.go
```diff
@@ -0,0 +1,134 @@
...
+	prefundedAccounts := make([]genesis.PrefundedAccount, 0)
+
+	switch accounts.(type) {
+	// Accounts come from config file.
+	case []interface{}:
+		accountArray := accounts.([]interface{})
```
jmatosp: ah, now I get it, maybe the comment wont be necessary if you extract each block of the case to it's own function

`prefundedAccountsFromConfig` and `prefundedAccountsFromCLI`?

```
switch accounts.(type) {
   case []interface{}:
      accountArray := accounts.([]interface{})
      prefundedAccountsFromConfig(accountArray)

   case string:
      accountsString := accounts.(string)
      prefundedAccountsFromCLI(accountsString)
}
```

kcoin/genesis/genesis_generator.go
jgimeno: I was having problems to manage big numbers from the console or the config files. All the values in the options are of type string, in this case we can benefit too different types of numbers for the configuration as decimal, hex, etc.

After it is converted to big int, address, etc

kcoin/genesis/genesis_generator_test.go
```diff
@@ -0,0 +1,300 @@
...
+				return command
+			},
+			ExpectedError: ErrInvalidConsensusEngine,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.TestName, func(t *testing.T) {
+			_, err := GenerateGenesis(test.InvalidCommandFromValid(baseValidCommand))
+			if err != test.ExpectedError {
```
jmatosp: `assert.EqualError(t, test.ExceptedError, err)`

I think this will log the case under test

kcoin/genesis/genesis_generator_test.go
```diff
@@ -0,0 +1,300 @@
...
+		PrefundedAccounts: []PrefundedAccount{
+			{
+				WalletAddress: "0xe2ac86cbae1bbbb47d157516d334e70859a1bee4",
+				Balance:       "15",
+			},
+		},
+	}
+
+	generatedGenesis, err := GenerateGenesis(cmd)
+
```
jmatosp: `assert.NoError(t, err)`

kcoin/genesis/genesis_generator_test.go
```diff
@@ -0,0 +1,300 @@
...
+	if err != nil {
+		t.Fatalf("Error: %s", err.Error())
+	}
+
+	fileName := "testfiles/testnet_default.json"
+	contents, err := ioutil.ReadFile(fileName)
```
jmatosp: `assert.NoError(t, err)` or `require.NoError(t, err)`

kcoin/genesis/genesis_generator_test.go
```diff
@@ -0,0 +1,300 @@
...
+	if err != nil {
+		t.Fatalf("Failed to read file %s", fileName)
+	}
+
+	var expectedGenesis = new(core.Genesis)
+	err = json.Unmarshal(contents, expectedGenesis)
+	if err != nil {
```
jmatosp: assert.NoError(t, err) or require.NoError(t, err)

kcoin/genesis/genesis_generator_test.go
```diff
@@ -0,0 +1,300 @@
...
+		},
+	}
+
+	generatedGenesis, err := GenerateGenesis(cmd)
+
+	if err != nil {
+		t.Fatalf("Error: %s", err.Error())
+	}
+
+	fileName := "testfiles/testnet_default.json"
```
jmatosp: yeah! golden files!

add a 

`var update = flag.Bool("update", false, "update .golden files")`

and then

```
if *update {
		t.Log("update golden file")
		if err := ioutil.WriteFile(gp, b.Bytes(), 0644); err != nil {
			t.Fatalf("failed to update golden file: %s", err)
		}
	}
```

so you can generate the golden file from tests when flag is set

`go test -v ./... -update`

more details
https://medium.com/soon-london/testing-with-golden-files-in-go-7fccc71c43d3

kcoin/genesis/genesis_generator_test.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		expectedExtradata := make([]byte, 32)
+		expectedExtradata = append([]byte(extraDataStr), expectedExtradata[len(extraDataStr):]...)
+
+		assert.Equal(t, expectedExtradata, generatedGenesis.ExtraData)
+	})
+}
+
+func TestItFailsWithAnInvalidNetwork(t *testing.T) {
```
jmatosp: isn't this a dup test from the beginning of the file, or just the name `TestMapNetwork`?

kcoin/genesis/genesis_generator_test.go
```diff
@@ -0,0 +1,300 @@
...
+		t.Run(test.testName, func(t *testing.T) {
+			_, err := mapNetwork(test.network)
+			if err != ErrInvalidNetwork {
+				t.Fatalf("Failed to throw exception with an invalid Network value.")
+			}
+		})
+	}
+}
+
+func TestItFailsWithInvalidConsensusEngine(t *testing.T) {
```
jmatosp: same here `TestMapConsensusEngine`?

kcoin/genesis/genesis_generator.go
jmatosp: :+1:

cmd/genesis/genesis.go
jgimeno: nice

cluster/genesis.go
```diff
@@ -836,10 +21,32 @@ func (client *cluster) generateGenesis() error {
 	if err != nil {
 		return err
 	}
-	var b bytes.Buffer
-	genesisTpl.Execute(&b, map[string]string{
-		"NetworkID": client.NetworkID,
-	})
+
+	newGenesis, err := genesis.GenerateGenesis(
+		genesis.GenesisOptions{
```
rgeraldes: genesis.Options vs genesis.GenesisOptions?

cluster/genesis.go
```diff
@@ -836,10 +21,32 @@ func (client *cluster) generateGenesis() error {
...
+			Network:                       "test",
+			MaxNumValidators:              "1",
+			UnbondingPeriod:               "0",
+			WalletAddressGenesisValidator: "0xd6e579085c82329c89fca7a9f012be59028ed53f",
```
rgeraldes: AccountAddressGenesisValidator

cluster/genesis.go
```diff
@@ -836,10 +21,32 @@ func (client *cluster) generateGenesis() error {
...
+			PrefundedAccounts: []genesis.PrefundedAccount{
+				{
+					WalletAddress: "0xd6e579085c82329c89fca7a9f012be59028ed53f",
```
rgeraldes: AccountAddress

cluster/genesis.go
```diff
@@ -836,10 +21,32 @@ func (client *cluster) generateGenesis() error {
...
+					Balance:       "0x200000000000000000000000000000000000000000000000000000000000000",
+				},
+				{
+					WalletAddress: "0x259be75d96876f2ada3d202722523e9cd4dd917d",
+					Balance:       "1000000000000000000",
+				},
+			},
+			SmartContractsOwner: "0x259be75d96876f2ada3d202722523e9cd4dd917d",
```
rgeraldes: This is probably going to be multisig: meaning that a contract will probably have multiple keys as the owner - please open an issue for this specific problem. We need to check this! :) For now we can leave it as is.

rgeraldes: SmartContractsOwner is too broad. I'd call it CoreContractsOwner

cluster/genesis.go
```diff
@@ -836,10 +21,32 @@ func (client *cluster) generateGenesis() error {
...
+		},
+	)
+
```
rgeraldes: extra space

cluster/minikube.go
```diff
@@ -77,7 +77,7 @@ func (cluster *minikubeCluster) Delete() error {
 
 // DockerEnv returns the environment variables necessary to connect to the private docker repository in the kubernetes cluster
 func (cluster *minikubeCluster) DockerEnv() ([]string, error) {
-	statusCmd := exec.Command("minikube", "docker-env", "-p", cluster.Name)
+	statusCmd := exec.Command("minikube", "docker-env", "-p", cluster.Name, "--shell", "bash")
```
rgeraldes: Had a quick look at the file and it seems that we are not checking errors for some cmd executions - not sure if its expected.

rgeraldes:
good job!

cluster/minikube.go
rgeraldes: Other: Also the internal/build/utils.go includes some utils for cmds. I'm not sure if its the best place though

acroca: @rgeraldes In all cmd executions we check for it's exit status to be 0. Where did you find a cmd without error checks?

cmd/genesis/genesis.go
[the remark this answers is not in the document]
acroca: I'd also prefer to either be specific with the data we're sending (if it's a string, use a separate func than the slice), or use always slices, even if there's only one element. Is that possible?

kcoin/genesis/genesis_generator_test.go
```diff
@@ -0,0 +1,305 @@
...
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"math/big"
+	"path/filepath"
+	"testing"
+)
+
+var update = flag.Bool("update", false, "update .golden files")
+
+func TestItFailsWhenRunningHandlerWithInvalidCommandValues(t *testing.T) {
```
acroca: Just a personal preference, I like to see positive tests first, errors/edge cases later. Main reason is to open a test file and the first thing to see is a kind of example of how this works, if you need more detail you can read more.

cluster/genesis.go
```diff
@@ -836,10 +21,32 @@ func (client *cluster) generateGenesis() error {
...
 		return err
 	}
-	var b bytes.Buffer
-	genesisTpl.Execute(&b, map[string]string{
-		"NetworkID": client.NetworkID,
-	})
+
+	newGenesis, err := genesis.GenerateGenesis(
+		genesis.GenesisOptions{
+			Network:                       "test",
```
acroca: You're not using the `client.NetworkID` anymore. We should change it to `client.Network` and use it here.

cluster/genesis.go
```diff
@@ -836,10 +21,32 @@ func (client *cluster) generateGenesis() error {
...
+			},
+			SmartContractsOwner: "0x259be75d96876f2ada3d202722523e9cd4dd917d",
+		},
+	)
+
+	if err != nil {
+		return err
+	}
+
+	out, _ := json.MarshalIndent(newGenesis, "", "  ")
```
acroca: Handle the error

kcoin/genesis/genesis_generator_test.go
jgimeno: Interesting point of view, never thought about it, has sense. I normally start with the edge cases than the happy path, that is why it appears in this order. I will change it.

cmd/genesis/genesis.go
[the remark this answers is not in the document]
acroca: if it overcomplicates things, then I'm fine with this. @jmatosp?

jgimeno:
@rgeraldes we include network in this case, that can be test or main. Do you want a specific   NetworkID/ChainID too? Is it needed?

PR #218, opened 2018-04-05
E2E test when an account doesn't have enough funds

acroca:
Added a test where a user doesn't have enough funds to make a transfer.


features/transactions.feature
```diff
@@ -9,5 +9,16 @@ I need to be able to send and receive transactions
       | A       | 10    |
       | B       | 5     |
     When I transfer 1 kcoin from A to B
```
yourheropaul: I propose we change the wording of the steps to a bit to describe the user's interactions and what they expect to see. I'd say what's here is pretty close already, it just needs a little bit of refinement.

The steps should be in natural language using terms the actor/persona (like, the user who is 'I' in this case) would recognise. They should be entirely from the perspective of the user (or sometimes users, in a network) and can ignore everything that the system is doing behind the scenes. For the purposes of this test specification, we care about _what_ the system does and not _how_ it does it. In other words, the user looks in at the system from the outside, and they just want it to work; it's the programmer's job to look after the inside of the system.

Basically cucumber is for creating product specifications and then using them for tests. It's not for designing test specifications (that's what godog is for).

The trick of getting the steps right is pretending you're talking the user through the scenario.  Users tend to be both annoyingly vague and quite specific. They might say, "I opened my laptop, started Chrome, went to the URL, and the page should be working" -- and you need to get details. Which OS? Which version of Chrome? Which URL, what does 'working' look like? When you have all the information, you can go through the scenario again with specific steps that are still in the user's language.

(It's not necessary to use the first person, or even the human perspective, in every step; as you've correctly identified in the existing features, sometimes the passive voice is more useful).

I would probably write the feature file as follows. (Gherkin has no comments as part of its syntax, so I've added some fake ones prefixed with `<--`).

```
Feature: Sending and receiving transactions  <-- I would put this is the present tense
As a technical user  <-- This user is quite technical, as we'll see
I want to be able to send and receive transactions

  Background: <-- This is for general network setup, in this case at least
    Given I am running a node on a network <-- as opposed to connecting via RPC
    And The network has 1 active validator
    And The network has 10 archive nodes
    
  Scenario: Send 1 kcoin successfully
    Given I have the following accounts:
      | account | funds |
      | A       | 10    |
      | B       | 5     |
    When I transfer 1 kcoin from A to B using the console <-- Explicit console mention 
    Then the balance of A should be around 9 kcoins <-- 'should be' because it reads naturally
    And the balance of B should be 6 kcoins <-- also 'kcoins' not 'kcoin'

  Scenario: Not enough funds
    Given I have the following accounts:
      | account | funds |
      | A       | 10    |
      | B       | 5     |
    When I try to transfer 11 kcoins from A to B using the console
    Then I should see a 'not enough funds' error <-- The exact error can be vague here; user doesn't care
    And the balance of A should be 10 kcoins
    And the balance of B should 5 kcoins
```

This is almost exactly the same in terms of functionality, but now it's a description of what the users wants to see and do, not what the system is doing. And yes, we do have to wait for the block to be processed in order to check the funds, but that's the nature of a blockchain.

yourheropaul:
As noted before, the tests are failing for me currently:

```
  Scenario: Send 1 kcoin                    # features/transactions.feature:6
    Given I have the following accounts:    # godogs_test.go:25 -> *Context
      | account | funds |
      | A       | 10    |
      | B       | 5     |
    When I transfer 1 kcoin from A to B     # godogs_test.go:26 -> *Context
    Then the last transaction is successful # godogs_test.go:29 -> *Context
    And the balance of A is around 9 kcoin  # godogs_test.go:28 -> *Context
    Balance expected to be around 9000000000000000000 but is 21000
    And the balance of B is 6 kcoin         # godogs_test.go:27 -> *Context
```

PR #231, opened 2018-04-10
RPC client for e2e tests client

### LINKED ISSUE #220: [e2e] Use an RPC client for e2e tests instead of Exec commands
### _Type_:
Request

I need to replace the `.Exec` commands with a RPC client connected to the cluster.

### _Priority_:
Low

acroca:
Close #220 

Introduced an RPC connection to the cluster from the e2e tests. At this point only the balance operations have been replaced to use the RPC connection instead of the `Exec`. More changes will come in a different PR.

Important changes:
- [x] Runs a kcoin pod and service with rpc enabled.
- [x] Balances use the RPC client instead of executing console commands


cluster/cluster.go
```diff
@@ -48,13 +48,38 @@ func (client *cluster) Cleanup() error {
...
+	for _, service := range list.Items {
+		err = client.Clientset.CoreV1().Services(Namespace).Delete(service.Name, &metav1.DeleteOptions{
+			GracePeriodSeconds: &zero,
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	return WaitFor(1*time.Second, 20*time.Second, func() bool {
```
jmatosp: isn't this function the same as the above?  refactor to `noServicesFound`?

acroca: They are very similar, one lists pods and the other one services...

acroca: will do, more readable

PR #233, opened 2018-04-12
Feature/e2e fast accounts

### LINKED ISSUE #227: [e2e] Create accounts without new instances
### _Type_:
Request

### _What do you need?_:

The e2e tests create a new node for each new account. We don't need that, we can just create an account without running anything in the cluster.

Doing so, it'll speed up the e2e tests run time.

### _Priority_:
Low

acroca:
Close #227 

Creating new accounts in the e2e tests doesn't instantiate a new instance for each account anymore. Instead, we have a key storage and create accounts there. With it, we can now send transactions using the client instead of console commands.

### DIFF
--- a/tests/features/accounts.go
+++ b/tests/features/accounts.go
@@ -60,6 +60,7 @@ func (ctx *Context) IHaveTheFollowingAccounts(accountsDataTable *gherkin.DataTab
 
 		ctx.accounts[accountData.AccountName] = account
 
+		// TODO: Use genesis validator key to send a transaction using the client instead of Exec
 		_, err = ctx.cluster.Exec(ctx.genesisValidatorName,
 			fmt.Sprintf(
 				`eth.sendTransaction({
--- a/tests/features/accounts.go
+++ b/tests/features/accounts.go
@@ -5,11 +5,9 @@ import (
 	"fmt"
 	"math/big"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/DATA-DOG/godog/gherkin"
-	"github.com/kowala-tech/kcoin/common"
 )
 
 type AccountEntry struct {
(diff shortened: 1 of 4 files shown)


tests/features/transactions.go
```diff
@@ -1,62 +1,41 @@
...
-			eth.sendTransaction({from:eth.coinbase, to: "%s", value: web3.toWei(%v, 'ether')})
-		`,
-		context.accountsCoinbase[to],
-		kcoin)
-	res, err := context.cluster.Exec(context.accountsNodeNames[from], command)
-	if err != nil {
-		return err
-	}
-	context.lastTxStdout = res.StdOut
+func (ctx *Context) ITryTransferKUSD(kcoin int64, from, to string) error {
```
jmatosp: err is always nil now

tests/features/accounts.go
```diff
@@ -48,47 +46,42 @@ func parseAccountsDataTable(accountsDataTable *gherkin.DataTable) ([]*AccountEnt
...
-			return err
-		}
-		coinbaseQuotes := res.StdOut
+		ctx.accounts[accountData.AccountName] = account
 
-		ctx.accountsNodeNames[account.AccountName] = nodeName
-		ctx.accountsCoinbase[account.AccountName] = strings.TrimSpace(strings.Replace(coinbaseQuotes, `"`, "", 2))
-
-		res, err = ctx.cluster.Exec(ctx.genesisValidatorName,
+		// TODO: Use genesis validator key to send a transaction using the client instead of Exec
```
jmatosp: move this information into a ticket? TODOs should not go in to stable branches

tests/features/transactions.go
acroca: Yes, but this is the step implementation. I'd prefer to keep it consistent with the rest, and return `error` even though right now it's always nil. Would you prefer to not return anything?

tests/features/accounts.go
acroca: Done

tests/features/transactions.go
acroca: Actually, godog handlers need to return `error`

jmatosp: Just notices, does need the err, it's godog method signature

PR #235, opened 2018-04-12
fixed warning on missing coinbase on validator start

### LINKED ISSUE #198: [node] warning should not be thrown - failed to get wallet account
### _Type_:
Issue

### _What do you need?_:
This warning should not be thrown as it confuses the user - 
WARN [04-03|10:39:47] failed to get wallet account             LOG15_ERROR=
This scenario happens when a key is not provided which is ok.

jmatosp:
Fixes to warn log message on validator start if no coinbase is set

Closes #198


.drone.yml
```diff
@@ -16,7 +16,7 @@ pipeline:
       - make lint
     when:
       event: [push]
-      branch: [feature/*, hotfix/*]
+      branch: [*]
```
yourheropaul: Is this right? Should we not add `bugfix/*` to the list rather?

kcoin/backend.go
```diff
@@ -319,24 +320,13 @@ func (s *Kowala) SetDeposit(deposit uint64) {
 }
 
 func (s *Kowala) StartValidating() error {
-	cb, err := s.Coinbase()
+	_, err := s.Coinbase()
 	if err != nil {
-		log.Error("Cannot start consensus validation without coinbase", "err", err)
+		log.Error("cannot start consensus validation without coinbase", "err", err)
```
yourheropaul: Is the case of error messages important?

jmatosp: you know go guy are very opinionated :)

it's up to us, but golang suggest that

https://github.com/golang/go/wiki/Errors

.drone.yml
jmatosp: we need to make sure that every build goes thru tests.

I notices that a branch named `issue` didn't went thru the CI test

we can add more branches prefixes

jmatosp: maybe remove the branch tag and change event to run on `pull_request`?

```
  test:
    image: kowalatech/go:1.0.1
    commands:
      - make test
      - make lint
    when:
      event: [pull_request]
```

jmatosp:
passing now, on fail the merge is blocked anyway

kcoin/backend.go
jmatosp: true, worse having mixed, should I revert or change this file?

.drone.yml
yourheropaul: Right, I see the problem. `event: [pull_request]` makes the most sense then, since every commit has to go through a PR.

kcoin/backend.go
yourheropaul: I suppose revert for now and make a do a complete sweep in a different PR?

PR #384, opened 2018-06-12
use validator methods in a safer way

### LINKED ISSUE #380: Fix validation methods
Some validation methods fail in the js console.

Open a console with `docker run --rm -it kowalatech/kusd:dev --testnet console`

- `validator.setDeposit(1)` fails
- `validator.getDeposits()` fails

/cc @rgeraldes @jmatosp @JekaMas

JekaMas:
Important changes:
- [x] `validator.getDeposits()`, `validator.redeemDeposits()`, `validator.setDeposit()` now return error in case if validator is not running
- [x] `validator.setDeposit()` uses `big.Int` so it can't unmarshal values like "0x0010101010" (hexutil.Big)
- [x] `validator.start()` receives an optional parameter `deposit big.Int`

Closes #380


internal/web3ext/web3ext.go
```diff
@@ -454,6 +454,8 @@ web3._extend({
 		new web3._extend.Method({
 			name: 'start',
 			call: 'validator_start',
+			params: 1,
+			inputFormatter: [null]
```
jmatosp: does it need the input formatter?

knode/api.go
```diff
@@ -57,14 +58,21 @@ func NewPrivateValidatorAPI(kcoin *Kowala) *PrivateValidatorAPI {
 }
 
 // Start the validator.
-func (api *PrivateValidatorAPI) Start() error {
+func (api *PrivateValidatorAPI) Start(deposit *big.Int) error {
```
jmatosp: do we really need this, it add a dependency on validator

knode/backend.go
```diff
@@ -320,12 +324,16 @@ func (s *Kowala) GetMinimumDeposit() (*big.Int, error) {
...
 
 // set in js console via admin interface or wrapper from cli flags
-func (s *Kowala) SetDeposit(deposit *big.Int) {
+func (s *Kowala) SetDeposit(deposit *big.Int) error {
 	s.lock.Lock()
 	s.deposit = deposit
 	s.lock.Unlock()
 
-	s.validator.SetDeposit(deposit)
+	if err := s.validator.SetDeposit(deposit); err != nil {
```
jmatosp: `return s.validator.SetDeposit(deposit)`

knode/backend.go
```diff
@@ -336,7 +344,7 @@ func (s *Kowala) StartValidating() error {
 	}
 
 	deposit, err := s.Deposit()
-	if err != nil {
+	if err != nil && err != validator.ErrIsNotRunning {
```
jmatosp: I think the start as a deferred, so not sure why this check

knode/validator/validator.go
```diff
@@ -191,8 +192,14 @@ func (val *validator) SetCoinbase(walletAccount accounts.WalletAccount) error {
 	return nil
 }
 
-func (val *validator) SetDeposit(deposit *big.Int) {
+func (val *validator) SetDeposit(deposit *big.Int) error {
+	if !val.Validating() {
```
jmatosp: this is not true, you can only setDeposit if NOT running

knode/validator/validator.go
```diff
@@ -642,9 +649,15 @@ func (val *validator) updateValidators(checksum [32]byte, genesis bool) error {
...
 
 func (val *validator) Deposits() ([]*types.Deposit, error) {
+	if !val.Validating() {
+		return nil, ErrIsNotRunning
+	}
 	return val.consensus.Deposits(val.walletAccount.Account().Address)
 }
 
 func (val *validator) RedeemDeposits() error {
+	if !val.Validating() {
```
jmatosp: same here

knode/validator/validator.go
```diff
@@ -642,9 +649,15 @@ func (val *validator) updateValidators(checksum [32]byte, genesis bool) error {
 }
 
 func (val *validator) Deposits() ([]*types.Deposit, error) {
+	if !val.Validating() {
```
jmatosp: same here

internal/web3ext/web3ext.go
JekaMas: `inputFormatter: [null]` stands for single optional parameter

knode/validator/validator.go
JekaMas: Old logic was `you can set deposit in any time`. For me, it's incorrect and leads to errors.

knode/api.go
acroca: I don't think we need it.
But if we keep this argument here, shall we get rid of the `setDeposit` and make the `deposit` mandatory?

JekaMas: It's an optional parameter. A user can start a validator with `validator.setDeposit(10); validator.start();` or `validator.start(10);`

knode/backend.go
JekaMas: This is needed to not to return `validator.ErrIsNotRunning` in `validator.start()`.

knode/backend.go
JekaMas: Ok, will fix.

knode/api.go
JekaMas: @acroca I like the idea of removing setDeposit. It can prevent a lot of errors.

jmatosp: :+1:

JekaMas: @acroca @jmatosp removed. I kept `validator.start([deposit])` with an optional parameter to handle scenario:
1. start with deposit 10
2. stop
3. start without the deposit parameter to use old value.

knode/validator/validator.go
JekaMas: If a validator is not validating the next line `return val.consensus.Deposits(val.walletAccount.Account().Address)` will panic. We can check deposits only if in validating state.

knode/validator/validator.go
jmatosp: this is confusing, can we remove all SetDeposit and just assume the Start NEEDs the deposit value

knode/validator/validator.go
jmatosp: I tough that deposits is just a contract that you ask for a list?

JekaMas: It's true but it uses `val.walletAccount.Account()` that is not `nil` only after `Start()`. How do you think, maybe we can make a new issue to rewrite this piece of code to use Deposits without Start pre-requirements?

knode/validator/validator.go
JekaMas: It uses `val.walletAccount` too in more details, this is described below.

knode/validator/validator.go
jmatosp: yeah, I guess, this is God like class

JekaMas: Done https://github.com/kowala-tech/kcoin/issues/392

rgeraldes: Any user should be able to check their deposits even if it's not validating.

JekaMas: I think so. I've rewritten `Deposits()` in the last commit.

JekaMas:
@rgeraldes 
getDeposits return an error if no address given by a user and validator.Start is not called.
validator.redeemDeposits() - returns an error if validator is not started. If we remove the check it'll panic. To handle that issue we have https://github.com/kowala-tech/kcoin/issues/392
setDeposit - removed.

The issue is caused by lack of checks in getDeposits and redeemDeposits. They both panics without this check.

rgeraldes:
What if a I want to check my deposits and don't want to start a validator? that's a valid use case.

knode/validator/validator.go
```diff
@@ -191,8 +193,14 @@ func (val *validator) SetCoinbase(walletAccount accounts.WalletAccount) error {
 	return nil
 }
 
-func (val *validator) SetDeposit(deposit *big.Int) {
+func (val *validator) SetDeposit(deposit *big.Int) error {
+	if val.Validating() {
```
rgeraldes: I think that this validation is not necessary anymore. we just call SetDeposit upon the validators' start method which verifies if the validator is running, or when we launch the client.

PR #387, opened 2018-06-13
export validator.isValidating

### LINKED ISSUE #386: export isValidating in web3.ext
### _Type_:
Request

`isRunning` can't say much about validator's state.

### _Priority_:
Medium

JekaMas:
Important changes:
- [x] adds `validator.isValidating()`

Closes #386

### DIFF
--- a/internal/web3ext/web3ext.go
+++ b/internal/web3ext/web3ext.go
@@ -496,6 +496,10 @@ web3._extend({
 			call: 'validator_isRunning'
 		}),
 		new web3._extend.Method({
+			name: 'isValidating',
+			call: 'validator_isValidating'
+		}),
+		new web3._extend.Method({
 			name: 'redeemDeposits',
 			call: 'validator_redeemDeposits'
 		}),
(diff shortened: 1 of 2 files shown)


internal/web3ext/web3ext.go
```diff
@@ -495,6 +495,10 @@ web3._extend({
 			name: 'isRunning',
 			call: 'validator_isRunning'
 		}),
+		new web3._extend.Method({
+			name: 'isValidating',
```
acroca: What's the usecase of `validator.isRunning` vs `validator.isValidating`? As a user of `validator`, when should I use either one? They sound similar, like if a validator object is running for me it means it's validating, that's why it's called validator... no?

JekaMas: Afaik validator can be in 2 states: running and validating.
If validator is 'running' it doesn't mean that it is 'validating'. But if it's validating, it's running also.
Maybe we should remove isRunning?

acroca: I don't know. I don't understand why a validator would be running and not validating. But if that's a thing, then it's alright.

jmatosp: I think running means that is trying to start validating. the user says start, but hasn't yet joined the validators

acroca: I see. Thanks @jmatosp

jmatosp: but, not easy to follow :|

PR #389, opened 2018-06-13
Feature/enable e2e ci

acroca:
Now we have all e2e tests in its own repo: https://github.com/kowala-tech/e2e

This PR removes unnecessary e2e code and enables the CI to run the e2e tests from that new repo.


acroca:
The e2e it's running well now, but the tests fail. We have to fix those tests first before merging this PR

/cc @JekaMas

Issue #360, opened 2018-06-07
Support multiple currencies


yourheropaul:
kUSD is the first currently, bu the goal is to support many different currencies (kEUR, kGBP,  kCDN, etc.). To accomplish this, we probably need a `--currency=kUSD` flag (with kUSD being the default for now) that sets up:

- [x]  The right genesis file (combined with `--testnet` flags)
- [x] Names of the currency (mostly in things like the console)
- [ ] Any oracle-related config
- [x] Data paths

jgimeno:
The oracle is confirmed with ricardo that there is nothing to do for now on this.

PR #402, opened 2018-06-18
Feature/issue 360 data paths currency

jgimeno:
Now it uses the currency as datapath.

Closes #
Last checkbox in https://github.com/kowala-tech/kcoin/issues/360


rgeraldes:
This PR should not be merged right away - requires multiple reviewers.

cmd/kcoin/chaincmd.go
```diff
@@ -157,8 +159,14 @@ func initGenesis(ctx *cli.Context) error {
 }
 
 //extractKeyCoin returns the keycoin used for launching the client. For now it always return kusd.
```
acroca: Update the comment.

cmd/kcoin/chaincmd.go
```diff
@@ -157,8 +159,14 @@ func initGenesis(ctx *cli.Context) error {
...
-func extractKeyCoin(context *cli.Context) string {
-	return "kusd"
+func extractKeyCoin(ctx *cli.Context) string {
+	kcoin := ctx.GlobalString(utils.CurrencyFlag.Name)
+
+	if kcoin == "" {
```
acroca: Its default value is `"kusd"`. Why do we need this if? Just for safety?

acroca:
Do we want to limit the values we accept for the currency or it's fine to keep it open? I don't see any harm in running it with an invalid currency, but perhaps someone will use `kUSD` by mistake instead of `kusd`...

cmd/kcoin/chaincmd.go
jgimeno: Yeah, in theory never will happen.

jgimeno:
@acroca for now it fails with an invalid currency, since it checks for the available ones when generating the genesis.

jgimeno:
Even kUSD

acroca:
awesome. Does it show a meaningful error message? :)

jgimeno:
Absolutely @acroca ;P 

It says invalid kcoin.

PR #412, opened 2018-06-20
Monorepo

acroca:
Moves everything in this repo into a subdirectory. This repo will become the monorepo to host all other projects, but this is the first step.

- Move codebase to `client/`
- Update imports to reflect the new path
- Fix the build processes to support one level deeper sourcecode
- Put Drone yaml in the root
- Put Makefile in the root


yourheropaul:
Drone expects `.drone.yml` to live in the root, which is why it's not processing the commit.

acroca:
For a review feel free to ignore the first two commits.
The first one just moves everything to `client`, the 2nd one is the result of:
```
find client -type f -not -path "client/vendor/*" -exec sed -i 's,github.com/kowala-tech/kcoin,github.com/kowala-tech/kcoin/client,g' {} +
```

yourheropaul:
The readme should also be in the root

acroca:
I actually moved there too.

PR #428, opened 2018-06-22
[418] function _hasAvailability() changed from public to private

JuliusBrain:
A short summary.

https://github.com/kowala-tech/kcoin/pull/418

Closes #


rgeraldes:
yes, the contracts must be regenerated.

JuliusBrain:
After running `go generate` inside `client/contracts/consensus/` I got following error.

`contracts/mgr/ValidatorMgr.sol:3:1: Error: Source "/usr/local/include/solidity/lifecycle/contracts/Pausable.sol" not found: File not found.
import "github.com/kowala-tech/kcoin/client/contracts/lifecycle/contracts/Pausable.sol" as pausable;
^--------------------------------------------------------------------------------------------------^
contracts/mgr/ValidatorMgr.sol:4:1: Error: Source "/usr/local/include/solidity/token/contracts/ERC223.sol" not found: File not found.
import "github.com/kowala-tech/kcoin/client/contracts/token/contracts/ERC223.sol" as token;
^-----------------------------------------------------------------------------------------^
consensus.go:16: running "solc": exit status 1`

client/contracts/token/contracts/Token.Sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity 0.4.21;
+pragma solidity ^0.4.21;
```
rgeraldes: This is usually a bad practise as we should use a concrete version

JuliusBrain: Why do think that using ^ is bad practice? It helps greatly with specifying the minimal version our code needs to be compliant and if we want to update solidity, we do not need to change every contract.

JuliusBrain: http://solidity.readthedocs.io/en/v0.4.24/layout-of-source-files.html#version-pragma

rgeraldes: https://github.com/ConsenSys/smart-contract-best-practices/blob/master/docs/recommendations.md
Section: Lock pragmas to specific compiler version

JuliusBrain: Ok, got it. Will change that.

rgeraldes:
Since you changed some method names you will need to update some references on the go code side. To check the project's integrity: run "make all" and "make test"

rgeraldes:
PR is deprecated  but the issue may still be valid - closing it for now.

PR #431, opened 2018-06-25
[423] Added truffle and npm to the project. Go wrappers for contracts…

JuliusBrain:
… and contracts itself are copied to goContracts for future refactor

A short summary.

Important changes:

Mandatory:
- [x] Added documentation.

Closes #


client/contracts/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "sharding",
```
yourheropaul: I think this package is probably from anther project

client/contracts/truffle.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  networks: {
+    ganache: {
+      host: '127.0.0.1',
```
yourheropaul: Ganache is a testing sandbox, if I remember correctly. Can this be used for integration tests?

yourheropaul:
This seems like an interesting approach! A Truffle-based deployment strategy for contracts (even in the testnets) could make everything much more visible.

I have some general questions though:

1. At the moment all the unit tests for the contracts are in Go, using the Go bindings. It seems like there are probably more standard (and possibly better) node-based Solidity testing tools. Is it easy or worthwhile to port the Go tests?

2. Is it easy to run those tests in the CI, in such a way that we get (ideally more) useful information about any issues? We use drone.io, which is Docker-based, so are there existing docker images with pre-installed Truffle tools (running npm install every time really slows things down)?

3. We're building end-to-end tests for everything in the system using a BDD-like system (and specifically [cucumber features](https://github.com/kowala-tech/kcoin/tree/dev/e2e/features)). That's possible because we can quickly Docker images (including compiled binaries with bespoke genesis blocks) and spin up ad hoc testnets with 1 second block times. Is there a good way to automatically and quickly deploy contracts into a system like that?

4. We're trying to enforce coding standards as part of the CI process. That's fairly easy in Go, because formatting is part of the syntax and there are really good [linting tools](https://github.com/alecthomas/gometalinter) that come with the language. Is there a good formatter, linter and preflight syntax checker for Solidity?

client/contracts/package.json
JuliusBrain: Yes, it was taken from the test project I've done during the interview process :) I will fix that

client/contracts/truffle.js
JuliusBrain: Yes, it can be used for integration tests because it simulates ethereum network. Ganache replaced testrpc in truffle 4.0 if I remember correctly

JuliusBrain:
Answering questions:

_1. At the moment all the unit tests for the contracts are in Go, using the Go bindings. It seems like there are probably more standard (and possibly better) node-based Solidity testing tools. Is it easy or worthwhile to port the Go tests?_
I think it is worthwhile to port Go tests as we will have unified place for smart contracts development and testing. Leaving some tests written in Go and have separate tests in truffle will make a mess later in the project. Maybe not now, but in the future when we will have more and more smart contracts, it will harder to maintain both of the tests. I wanted to rewrite Go tests for truffle

_2.Is it easy to run those tests in the CI, in such a way that we get (ideally more) useful information about any issues? We use drone.io, which is Docker-based, so are there existing docker images with pre-installed Truffle tools (running npm install every time really slows things down)?_
I can search for those but I think any docker image with npm pre-installed will be enough. In my previous job, we used Buitbucket pipelines (CI) and we run npm install everytime for much more smart contracts (whole ICOs) and it didn't slow things down, checking maybe took (with tests running also) maybe 2 minutes max.

_3. We're building end-to-end tests for everything in the system using a BDD-like system (and specifically cucumber features). That's possible because we can quickly Docker images (including compiled binaries with bespoke genesis blocks) and spin up ad hoc testnets with 1 second block times. Is there a good way to automatically and quickly deploy contracts into a system like that?_
In my understanding of truffle, you can specify any network in a truffle config and run migration to that network. 

_4.We're trying to enforce coding standards as part of the CI process. That's fairly easy in Go, because formatting is part of the syntax and there are really good linting tools that come with the language. Is there a good formatter, linter and preflight syntax checker for Solidity?_
Yes, I already added them into the package.json file. Eslint for truffle tests and Solium for Solidity. There are scripts ready in the package.json file. We can specify our own rules as with every linter :)

PR #434, opened 2018-06-26
Feature/add gource

adamking:
Add gource video and compile/upload on push to dev

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -122,6 +122,7 @@ pipeline:
 
 
   build-artifacts:
+    group: build-artifacts
     image: kowalatech/go:1.0.3
     environment:
       - DOCKER_HOST=tcp://docker:2375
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -107,6 +107,28 @@ pipeline:
       event: [push, tag]
       branch: [dev]
 
+  compile-gource-video:
```
yourheropaul: Could you put this in a group (along with `build-artefacts`) so that is builds in parallel?

.drone.yml
```diff
@@ -107,6 +107,28 @@ pipeline:
...
+    repo: kowalatech/kcoin
+    secrets: [ docker_username, docker_password ]
+    dockerfile: gource/Dockerfile
+    commands:
+      - make gource-video
+    when:
+      branch: [dev]
+      event: [push]
+
+  upload-gource-mp4-to-s3:
```
yourheropaul: Could you put this in a group with the other s3 deployment pipeline?

gource/.gitignore
```diff
@@ -0,0 +1,2 @@
+results/*
```
yourheropaul: These may as well go in the master .gitignore in the root

.drone.yml
```diff
@@ -107,6 +107,28 @@ pipeline:
       event: [push, tag]
       branch: [dev]
 
+  compile-gource-video:
+    repo: kowalatech/kcoin
+    secrets: [ docker_username, docker_password ]
+    dockerfile: gource/Dockerfile
```
yourheropaul: The pipeline will need an `image:` tag, which is why the CI is complaining. To make it work, we need a docker image on DockerHub. If there isn't one already, then the gource/Dockerfile should be split out to a different repo and pushed to dockerhub on commit.

.drone.yml
```diff
@@ -107,6 +107,28 @@ pipeline:
...
+    commands:
+      - make gource-video
```
yourheropaul: This make task uses docker, which isn't necessary when the pipeline uses a gource docker image. I guess there needs to be a new make task that just runs `gource`, and assumed that everything is installed to make that happen (which should be the case because it will run inside an appropriate docker image in the CI).

PR #443, opened 2018-06-27
fetch vendor in the dockerfile

acroca:
Build was failing because vendor wasn't fetched. This makes the dockerfile fetch all vendor source.


acroca:
I have to leave now, please approve and merge for me, it'll fix the build

Issue #490, opened 2018-07-06
[netstats and explorer] Include new stats


yourheropaul:
Now that we have improved stats coverage, we should include it in the right places.

yourheropaul:
+    MinDeposit     *big.Int `json:"minDeposit"`
+    ValidatorCount *big.Int `json:"validatorCount"`
+    MaxValidators  *big.Int `json:"maxValidators"`
+    OracleCount    *big.Int `json:"oracleCount"`
+    CurrencyPrice  *big.Int `json:"currencyPrice"`

PR #511, opened 2018-07-10
Netstats

### LINKED ISSUE #462: [monorepo] Netstats
https://github.com/kowala-tech/kUSD-netstats

@acroca

yourheropaul:
Caveats:

- We currently don't know the total number of registered ~validators~ oracles
- We currently don't know the stability fee
- I've restructured the stats output from the client, so we'll have to restart the network for the new stats to take effect

Closes #462


rgeraldes:
"We currently don't know the total number of registered validators" - By active I was thinking about the ones that submitted the price, and by registered it's the value that we are providing which is the list of registered oracles of the contract.

yourheropaul:
I meant 'oracles' above, not 'validators'. The contract bindings and the stat paylods don't currently contain that information, so I didn't include it in the netstats. Issue #490 addresses this problem.

PR #518, opened 2018-07-12
Put e2e logs on S3

acroca:
When a build fails, it will send the e2e logs to S3. I set the verbosity to 6 for a more extensive logs with the risk of being too noisy.


acroca:
@JekaMas you need access to our S3 account

jmatosp:
@acroca did you had a look into shipping to our log aggregate?

acroca:
Maybe we should, I was looking at the logs of the build and using S3 is not very convenient.

yourheropaul:
Shipping to logz.io might be easier

acroca:
Do you guys mind shiping this first and move to something else later? It'll help fixing the current random erros

PR #545, opened 2018-07-13
client/knode/genesis: pre-funded the oracle's account - test network

rgeraldes:


yourheropaul:
@rgeraldes There are broken unit tests in this PR.

Issue #467, opened 2018-07-02
[CI] include file generation (go generate) during the build


rgeraldes:
### _Type_:
Issue

### _What happened?_:
The file generation was failing but it's not visible during the build since the codebase relies on the available generated files which were created on previous snapshots of the codebase that did not contain errors. We must include file generation in the build process to identify those situations and to make sure that we are using the latest version.

### _Affected servers or services_:
Contract bindings

### _Priority_:
High

jgimeno:
@rgeraldes is this already done? I see on the  make bindings:

```# FILES is the list of binding files that would be created when generating the bindings
FILES=$(shell egrep -ir "go:generate" client/contracts/bindings | grep abigen | sed -E 's/^client\/contracts\/bindings\/(.*)\/.*\.go.*-out\ \.?\/?(.*)/client\/contracts\/bindings\/\1\/\2/' )
$(FILES):
	$(MAKE) -j 5 stringer go-bindata gencodec client/build/bin/abigen client/contracts/truffle/node_modules
	go generate ./client/contracts/bindings/...
bindings: | $(FILES)```

PR #559, opened 2018-07-17
Feature/issue 467 go generate on build

jgimeno:
Now the build checks that the go generated files are up to date, if not it fails.

Closes #

https://github.com/kowala-tech/kcoin/issues/467


Makefile
```diff
@@ -92,6 +92,38 @@ client/contracts/truffle/node_modules:
...
+	go generate ./client/internal/jsre/deps/
+	go generate ./client/knode/
+	go generate ./client/knode/tracers/internal/tracers/
+	go generate ./client/p2p/discv5/
+	go generate ./notifications/blockchain/
+	go generate ./notifications/environment/
+	go generate ./notifications/keyvalue/
+	go generate ./notifications/notifier/
+	go generate ./notifications/protocolbuffer/
+	go generate ./wallet-backend/protocolbuffer/
```
acroca: I think we'll easily forget to keep this list up to date. What if we do like in the `bindings` task?

Something like:
```
egrep -ir --exclude-dir=vendor --exclude-dir=_workspace --include \*.go "//go:generate" . | grep -v client/contracts | sed -E 's/^([^\:]*).*$/\1/' | uniq
```

Makefile
```diff
@@ -92,6 +92,38 @@ client/contracts/truffle/node_modules:
 client/build/bin/abigen:
 	cd client; build/env.sh go run build/ci.go install ./cmd/abigen
 
+go-generate: moq go-bindata stringer gencodec mockery ensure-notifications ensure-wallet-backend
+	go get -u github.com/golang/protobuf/protoc-gen-go
```
acroca: Move this to its own task at the end (with the other tools) and only install it if missing

.drone.yml
```diff
@@ -10,13 +10,19 @@ clone:
 
 pipeline:
   bindings:
-    image: kowalatech/go:1.0.4
+    image: kowalatech/go:1.0.8
     commands:
       - make bindings
 
+  go-generate:
```
acroca: I'd make this run on `push` only

Makefile
```diff
@@ -92,6 +92,38 @@ client/contracts/truffle/node_modules:
...
+	go generate ./client/p2p/discv5/
+	go generate ./notifications/blockchain/
+	go generate ./notifications/environment/
+	go generate ./notifications/keyvalue/
+	go generate ./notifications/notifier/
+	go generate ./notifications/protocolbuffer/
+	go generate ./wallet-backend/protocolbuffer/
+
+assert-no-generate:
+	git status
```
acroca: Is this just to give some context to the error or a mistake?

Makefile
```diff
@@ -92,6 +92,38 @@ client/contracts/truffle/node_modules:
...
+	if ! git diff-index --quiet HEAD; then echo "There are uncommited go generate files."; exit 1; fi
+
+ensure-notifications: dep
+	cd notifications && \
+	$(GOPATH)/bin/dep ensure --vendor-only && \
+	cd ..
```
acroca: I think you don't need to go back

Makefile
jgimeno: I prefer leaving this status just for getting context, when it fails it is not always easy to know what happened. So it is not harmful to have this.

Makefile
```diff
@@ -92,6 +92,38 @@ client/contracts/truffle/node_modules:
 client/build/bin/abigen:
 	cd client; build/env.sh go run build/ci.go install ./cmd/abigen
 
+go-generate: moq go-bindata stringer gencodec mockery ensure-notifications ensure-wallet-backend
```
acroca: The makefile tasks have to use underscores, so `go_generate`

Makefile
```diff
@@ -92,6 +92,38 @@ client/contracts/truffle/node_modules:
...
+	go generate ./client/knode/tracers/internal/tracers/
+	go generate ./client/p2p/discv5/
+	go generate ./notifications/blockchain/
+	go generate ./notifications/environment/
+	go generate ./notifications/keyvalue/
+	go generate ./notifications/notifier/
+	go generate ./notifications/protocolbuffer/
+	go generate ./wallet-backend/protocolbuffer/
+
+assert-no-generate:
```
acroca: underscores

Makefile
```diff
@@ -92,6 +92,38 @@ client/contracts/truffle/node_modules:
...
+	git status
+	if ! git diff-index --quiet HEAD; then echo "There are uncommited go generate files."; exit 1; fi
+
+ensure-notifications: dep
```
acroca: underscores

Makefile
```diff
@@ -92,6 +92,38 @@ client/contracts/truffle/node_modules:
...
+	cd notifications && \
+	$(GOPATH)/bin/dep ensure --vendor-only && \
+	cd ..
+
+ensure-wallet-backend: dep
```
acroca: underscores

Makefile
acroca: actually, if we combine this with the bindings, we can just `go generate ./...`

jgimeno: go generate fails if bindings dont happen before the other, at least this was what I was getting on the beginning.

jgimeno: I have been trying the approach you gave me, but the egrep version of the ci does not have that options. If you can help me in this case? I am not as good in bash!

@acroca

PR #565, opened 2018-07-18
feature/super node status

rgeraldes:
- super node status logic (added config parameter: superNodeAmount)
- oracleMgr: included registration modifier - verifies if the user is a super node
- turned the test network genesis validator into a super node (6000000 deposit)
- included tests (based on the go bindings)


yourheropaul:
The unit tests are failing on this. They're failing so hard that the CI isn't tell us.

PR #569, opened 2018-07-19
test for genesis block generated

jmatosp:
Test without panic on genesis block generated


client/knode/genesis/generator_test.go
```diff
@@ -3,16 +3,19 @@ package genesis
...
 	"testing"
+	"github.com/stretchr/testify/require"
 )
 
 func TestGenerateIsDeterministic(t *testing.T) {
 	options := Networks["kusd"][MainNetwork]
 
 	generatedGenesis, err := Generate(options)
 	assert.NoError(t, err)
+	require.NotNil(t, generatedGenesis)
```
acroca: Shouldn't the `Generate` return an actual object if there's no error?

jmatosp: this would test exactly that, but i guess I can require on the NoError instead

acroca: yup, that's what I thought. It's fine to have this doublechecked though...

PR #586, opened 2018-07-24
Added proxy contracts

JuliusBrain:
A short summary.

Important changes:

Mandatory:
- [x] Added documentation.

Closes #


client/contracts/truffle/contracts/kns/KNSRegistry.sol
```diff
@@ -1,10 +1,12 @@
 pragma solidity 0.4.24;
 
 import "./KNS.sol";
+import "../proxy/Initializable.sol";
+
 /**
  * The KNS registry contract.
  */
-contract KNSRegistry is KNS {
+contract KNSRegistry is KNS, Initializable {
```
jgimeno: If we have the V1 that supports proxy we can remove this one?

client/contracts/truffle/contracts/kns/KNSRegistryV1.sol
```diff
@@ -0,0 +1,108 @@
...
+    // Permits modifications only by the owner of the specified node.
+    modifier only_owner(bytes32 node) {
+        require(records[node].owner == msg.sender);
+        _;
+    }
+
+    /**
+     * @dev Constructs a new KNS registrar.
+     */
+    constructor() public {
```
jgimeno: Same here, is it needed the normal constructor?

jgimeno: Now I see that some of this contracts are needed for the only purpose of testing, can it be?

client/contracts/truffle/contracts/kns/PublicResolver.sol
```diff
@@ -57,6 +58,14 @@ contract PublicResolver {
         kns = knsAddr;
     }
 
+    /**
+     * @dev initialize function for Proxy Pattern.
+     * @param knsAddr The address of the KNS registry.
+     */
+    function initialize(KNS knsAddr) isInitializer public {
```
jgimeno: Same with both constructor and initialize, do we need both or just initialize?

client/contracts/truffle/test/proxy/Proxy.test.js
```diff
@@ -0,0 +1,91 @@
...
+const FIFSRegistrar = artifacts.require('FIFSRegistrar.sol');
+const PublicResolver = artifacts.require('PublicResolver.sol');
+const ValidatorMgr = artifacts.require('ValidatorMgr.sol');
+// const DummyImplementation = artifacts.require('DummyImplementation');
+const UpgradeabilityProxy = artifacts.require('UpgradeabilityProxyFactory.sol');
+const AdminUpgradeabilityProxy = artifacts.require('AdminUpgradeabilityProxy.sol');
+const namehash = require('eth-ens-namehash');
+
+contract('Proxy Functionality', ([_, admin, owner, anotherAccount]) => {
+  // before(async () => {
```
jgimeno: Commented code can be removed I guess.

client/contracts/truffle/contracts/kns/KNSRegistry.sol
JuliusBrain: From my understanding we need this as the next contract may also be upgraded, so it need to be Initializable. I will test if only the original contract needs to be Initializable

client/contracts/truffle/contracts/kns/FIFSRegistrar.sol
```diff
@@ -25,6 +26,16 @@ contract FIFSRegistrar {
         rootNode = node;
     }
 
+    /**
+     * @dev initialize function for Proxy Pattern.
+     * @param knsAddr The address of the KNS registry.
+     * @param node The node that this registrar administers.
+     */
+    function initialize(KNS knsAddr, bytes32 node) isInitializer public {
```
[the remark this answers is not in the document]
JuliusBrain: I tried removing constructor but got error "VM Error, transaction revert". After rest of the work is done, I will get back to this.

client/contracts/truffle/test/proxy/Proxy.test.js
JuliusBrain: Yes :)

client/contracts/truffle/contracts/kns/FIFSRegistrar.sol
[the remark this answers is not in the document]
jgimeno: oh! Good to know!

yourheropaul:
@jgimeno the generation process was recently updated, which may be a cause of the difference. You could diff the generated go files. I don't think it matters so much though

yourheropaul:
@JuliusBrain Would you please make the CI run the truffle tests? I don't think of of them are included at the moment, so we don't have regression protection.

PR #587, opened 2018-07-24
reverted incompatible dependency

### LINKED ISSUE #576: Netstats has unreliable websockets
Since it was added to the mono repo, the netstats application is having websocket troubles. Connections keep dropping.

I [updated some of its dependencies](https://github.com/kowala-tech/kcoin/commit/9082c160164a63aa9f87a2c7010600aa10b67ebb), which may be the cause. The [old, standalone version](https://github.com/kowala-tech/kUSD-netstats) seems to work.

jmatosp:
reverted falling dependency 

Closes #576

### DIFF
--- a/netstats/package-lock.json
+++ b/netstats/package-lock.json
@@ -97,9 +97,19 @@
       "integrity": "sha1-sqRdpf36ILBJb8N2jMJ8EvqRan0="
     },
     "async": {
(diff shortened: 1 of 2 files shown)


jmatosp:
Tested all dependencies changed and this is the offending one.

PR #597, opened 2018-07-30
Put commit and tag binary releases in folders

acroca:
For a cleaner S3 folder structure, puts commit and tag releases in folders.
The stable and unstable releases don't change, so all links pointing to them will still work.


yourheropaul:
That's an improvement. I'm going to merge it.

PR #606, opened 2018-08-01
Initial licenses

### LINKED ISSUE #605: Add License to the smart contracts
### _Type_:
Issue

We are missing licence files for our smart contracts. Some contracts, like KNS, are based on different contracts and needs to inherit their licence.

### _Priority_:
Medium

  * High = We should stop anything we are doing and do this.
If you're writing this calmly, you should probably choose another option.
If it's really that urgent, please make sure we read this.
  * Medium = It's important that we do this within a few days.
  * Low = We will consider this on our next sprint planning.

(Please be aware that your priority may not match ours, we'll use this as guidance only).

JuliusBrain:
A short summary.
Licensing for smart contracts

Important changes:

Mandatory:
- [x] Added documentation.

Closes #605


yourheropaul:
Quoth John:

>This looks really good. I would change `Users cannot use the contract or modified version of it in another blockchain` to `Users cannot use the source or a modified version of it in blockchain other than kUSD Mainnet`

Issue #519, opened 2018-07-12
Duplicate vote and reorg chain errors during the tests


JekaMas:
### _Type_:
Issue

### _What happened? / What do you need?_:
At the moment we ignore all errors in `client/knode/validator/validator.go` on `val.votingSystem.Add(addressVote)`.

One of the errors is a duplicate vote on prevote and vote stage:
```
INFO [07-12|09:24:16.428] Pre vote sub-election 
ERROR[07-12|09:24:16.430] a duplicate vote in voting table 0. Error: 
	Vote(ba8890175554d0f62551c453b1fdb988434c1780dc16675c2ea662599dc34ef6)
	Block Number:		30
	Block Hash:			0000000000000000000000000000000000000000000000000000000000000000
	Round:	  			0
	Type: 				0
	V:        			0x27
	R:        			0xf84e4561a704dd81237ce18cacd0defae2020228bf240df0f5a9780a555079
	S:        			0x79b753078f4bdaa92eaa0d9cfec8d5697ce237862757b70da1b07bee18edbaab
	Hex:      			f866a000000000000000000000000000000000000000000000000000000000000000001e8080279ff84e4561a704dd81237ce18cacd0defae2020228bf240df0f5a9780a555079a079b753078f4bdaa92eaa0d9cfec8d5697ce237862757b70da1b07bee18edbaab
 
WARN [07-12|09:24:16.430] Failed to add own vote to voting table   err="duplicate vote"
INFO [07-12|09:24:16.430] Waiting for a majority in the pre-vote sub-election 
INFO [07-12|09:24:16.430] There's a majority in the pre-vote sub-election! 
INFO [07-12|09:24:16.430] Pre commit sub-election 
ERROR[07-12|09:24:16.432] a duplicate vote in voting table 1. Error: 
	Vote(a7f78f200990c886e8ac306cc3dbab553174433cae50d1b408882744ce61a119)
	Block Number:		30
	Block Hash:			0000000000000000000000000000000000000000000000000000000000000000
	Round:	  			0
	Type: 				1
	V:        			0x27
	R:        			0xc5e163dc0ba10673f55274cd972aedb510ab49546df1f2b5bc0058390940eb96
	S:        			0x3703c773a58acf4983738905ab050f19da2f146c7beef21ec19bc144be0b06e7
[...]

PR #613, opened 2018-08-02
Validation sync errors

### LINKED ISSUE #612: Fix validation errors
### _Type_:
Issue

Validator tests often fail with errors:
```
  Scenario: Re-Start mining
    Given I have the following accounts:
      Error: timeout error: node is synchronised
```

```
--- Failed steps:

  Scenario: Re-Start mining
    And I wait for my node to be synced
      Error: timeout error: timeout waiting for node sync
```

Run validator tests many times

### _Priority_:
Medium

JekaMas:
The first iteration to make the validation process error-free.

Closes #612 https://github.com/kowala-tech/kcoin/issues/519


client/core/types/chunk.go
```diff
@@ -108,29 +113,49 @@ func (ds *DataSet) Size() uint {
...
 	// exists
+
+	ds.l.RLock()
+	defer ds.l.RUnlock()
 	return ds.data[i]
 }
 
-func (ds *DataSet) Add(chunk *Chunk) {
+func (ds *DataSet) Add(chunk *Chunk) error {
+	if chunk == nil {
```
acroca: I like this as a safety measure. But would be nice as well to detect when this situation happened and avoid it too.

client/core/types/vote.go
```diff
@@ -217,3 +218,39 @@ func (vote *Vote) String() string {
...
+	return &VotesSet{m: make(map[common.Hash]*Vote)}
+}
+
+func (v *VotesSet) Add(vote *Vote) {
+	v.l.Lock()
+	v.m[vote.Hash()] = vote
+	v.l.Unlock()
+}
+
+func (v *VotesSet) Has(h common.Hash) bool {
```
acroca: I'd rename it to `ContainsHash`.
Similar data structures use `Contains` instead of `Has` to check if an item is in the collection.
Also adding the `Hash` part is because his is a `VoteSet`, so I'd assume a `Contains` function would receive a `Vote` to check if it exists, so if we're checking its hash I think it's clearer to name it that way.

client/core/types/vote.go
```diff
@@ -217,3 +218,39 @@ func (vote *Vote) String() string {
 }
 
 type Votes []*Vote
```
acroca: Is the `VoteSet` a replacement for `Votes` or we still use both?

client/core/types/vote.go
```diff
@@ -217,3 +218,39 @@ func (vote *Vote) String() string {
...
+
+type VotesSet struct {
+	m map[common.Hash]*Vote
+	l sync.RWMutex
+}
+
+func NewVotes() *VotesSet {
```
acroca: `NewVoteSet` better, no?

client/core/voting_table.go
```diff
@@ -31,22 +31,25 @@ func NewVotingTable(voteType types.VoteType, voters types.Voters, majority Quoru
...
+		votes:    types.NewVotes(),
 		quorum:   TwoThirdsPlusOneVoteQuorum,
 		majority: majority,
 	}, nil
 }
 
-func (table *votingTable) Add(vote types.AddressVote) error {
-	if !table.isVoter(vote.Address()) {
-		return fmt.Errorf("voter address not found in voting table: %#x", vote.Address().Hash())
+func (table *votingTable) Add(voteAddressed types.AddressVote) error {
```
acroca: you could call it `addressVote`, maps better with `types.AddressVote` :)

client/knode/validator/validator.go
```diff
@@ -625,7 +630,7 @@ func (val *validator) AddBlockFragment(blockNumber *big.Int, round uint64, fragm
 		err = val.chain.Validator().ValidateState(block, parent, val.state, receipts, usedGas)
 		if err != nil {
 			val.handleMutex.Unlock()
-			log.Crit("Failed to validate the state", "err", err)
+			log.Error("Failed to validate the state", "err", err)
```
acroca: I'm worried about all those `Crit` to `Error`. I guess they were `Crit` because the state of the current process is invalid so we have to crash the process. Moving it to error means the process will continue. Is it fine? Will the state get fixed?

client/core/voting_table.go
JekaMas: I'm logging the request in https://github.com/kowala-tech/kcoin/pull/613/files#diff-658294a439327f8305b9d63bf426c417R581

client/core/types/vote.go
JekaMas: I do like `Contains(vote *Vote)` to hide the imlementation

client/core/types/vote.go
JekaMas: We still use both. `Votes` is used where we write new data to the commit to the DB.

client/core/types/chunk.go
JekaMas: I'm logging the request in https://github.com/kowala-tech/kcoin/pull/613/files#diff-658294a439327f8305b9d63bf426c417R581

client/core/types/vote.go
JekaMas: Sure.

client/knode/validator/validator.go
```diff
@@ -578,20 +578,27 @@ func (val *validator) vote(vote *types.Vote) {
...
+	if err := val.blockFragments.Add(fragment); err != nil {
+		err = errors.New("Failed to add a new block fragment: " + err.Error())
+		return err
+	}
 
 	if val.blockFragments.HasAll() {
 		block, err := val.blockFragments.Assemble()
 		if err != nil {
-			log.Crit("Failed to assemble the block", "err", err)
+			err = errors.New("Failed to assemble the block: " + err.Error())
```
JekaMas: @acroca I'll keep `Error` here because `Assemble` doesn't change DB state and using `Crit` leads to the possibility to kill any node in the Network just sending it an incorrect block fragment.

client/core/types/vote.go
acroca: Ok

PR #623, opened 2018-08-06
Replace binary with latest

jmatosp:


client/version/parser_test.go
```diff
@@ -17,6 +17,7 @@ func TestFilenameParser(t *testing.T) {
 		{"kcoin-1.0.0-windows-4.0-amd64.zip", "1.0.0", "windows-4.0", "amd64"},
 		{"kcoin-1.0.0-windows-4.0-amd64.exe.zip", "1.0.0", "windows-4.0", "amd64"},
 		{"kcoin-1.0.0-osx-10.6-amd64.zip", "1.0.0", "osx-10.6", "amd64"},
```
acroca: We don't use this name
Also, shouldn't we add the filenames we use now? For example, the windows release doesn't have the `-4.0` part, and the OSX doesn't have the `-10.6`

client/version/updater.go
```diff
@@ -54,13 +54,28 @@ func (u *updater) Update() error {
 	}
 
 	err = u.unzip()
+	if err != nil {
```
acroca: could use the `if err := u.unzip(); err != nil {` syntax instead. For this and for the following function calls.

client/version/parser_test.go
jmatosp: tests are outdated 
will change

PR #651, opened 2018-08-15
added error return from state func

### LINKED ISSUE #299: [validator] add block fragment panics
### _Type_:
Issue

Adding a block fragment panic if fails in multiple paths

`validator.go` line 577, 592 and 607

We should handle this errors gracefully, either withdraw from election  of exit validation process
but not panic the client

### _Priority_:
Medium

jmatosp:
Removing all panics from validator

Closes #299


client/knode/validator/validator.go
```diff
@@ -160,7 +160,14 @@ func (val *validator) run() {
 
 	log.Info("Starting the consensus state machine")
 	for state, numTransitions := val.notLoggedInState, 0; state != nil; numTransitions++ {
```
jgimeno: Is this valid? before it only returned one param notLoggedInState, but now it returns 2. Will it work???

jmatosp: hm, strange apparently does work, but your right.

Maybe has unexpected behaviour, let me do more testing

client/knode/validator/states.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/kowala-tech/kcoin/client/core/types"
 	"github.com/kowala-tech/kcoin/client/log"
 	"github.com/kowala-tech/kcoin/client/params"
+
```
rgeraldes: extra space

PR #672, opened 2018-08-22
knode/validator: transaction - wait until mined

### LINKED ISSUE #160: [consensus] replace isValidator with tx status

rgeraldes:
Closes #160


client/knode/validator/validator.go
```diff
@@ -700,5 +711,18 @@ func (val *validator) RedeemDeposits() error {
 	if !val.Validating() {
 		return ErrIsNotRunning
 	}
-	return val.consensus.RedeemDeposits(val.walletAccount)
+
+	txHash, err := val.consensus.RedeemDeposits(val.walletAccount)
+	if err != nil {
+		return err
+	}
+	receipt, err := tx.WaitMined(context.TODO(), val.backend, txHash)
```
JekaMas: How do you think, is it better to use here context.WithTimeout?

client/common/tx/confirmation.go
```diff
@@ -0,0 +1,40 @@
...
+	defer queryTicker.Stop()
+
+	logger := log.New("hash", txHash)
+	for {
+		receipt, err := b.TransactionReceipt(ctx, txHash)
+		if receipt != nil {
+			return receipt, nil
+		}
+		if err != nil {
+			logger.Trace("Receipt retrieval failed", "err", err)
```
JekaMas: 'return nil, err' is needed, I think

client/knode/validator/validator.go
rgeraldes: It needs to be TODO because if there was no error posting the tx the user might risk timing out (edge case) while the tx is successful which could lead to have their funds slashed if he does not participate in the consensus when he is supposed too. If we have a time out the user is responsible for checking those edge cases which I feel is risky.

PR #700, opened 2018-08-30
Improved console minting tools

yourheropaul:
Building on #664, the improves the console minting process.

## API

The basic calls are the same as before: one governance key submits a transaction, resulting in a _minting ID_, and another governance key confirms the transaction by referencing the minting ID.

#### Syntax

 `mtoken.mint(governance_key_address, recipient_address, amount_of_tokens)`

#### Example usage in the console 

```
> mtoken.mint(eth.coinbase, "0x1f1f1480f77b2565ae7f3a5580fd3da79b59b09b", 10); 
```

This would submit a transaction of 10 mtokens to the address ending `b09b`. The transaction will not be actioned until it's confirmed by another governance key.

#### Syntax

`mtoken.mintList()`

#### Example usage in the console

```
> mtoken.mintList()
```

For which the output might be something like:

```
[{
    amount: 42,
    confirmed: false,
    id: 0,
    to: "0x1f1f1480f77b2565ae7f3a5580fd3da79b59b09b"
}, {
    amount: 44,
    confirmed: true,
    id: 1,
    to: "0x1f1f1480f77b2565ae7f3a5580fd3da79b59b09b"
}]
```

The `id` field is the minting ID, which will be needed in the confirmation step. Once a transaction is submitted, it will appear at the end of this list for verification. The `confirmed` field indicated whether or not another governance key has confirmed the transaction (and thus actioned it).

#### Syntax

 `mtoken.confirm(governance_key_address, minting_id)`

#### Example usage

```
> mtoken.confirm(eth.coinbase, 1)
```

This would confirm the transaction with minting ID `1`, and action it. In the example output above, it would issue 42 tokens to the address ending `b09b`.


yourheropaul:
I feel like the implementation of this is a bit ugly, but I don't think there's a better way.

jmatosp:
Not sure why it's ugly, you mean from solution or code style?

yourheropaul:
I feel like there's a more canonical way of unpacking the data hashes. But it works, so I guess it's fine.

client/knode/api.go
```diff
@@ -256,6 +259,66 @@ func (api *PublicTokenAPI) MintingFinished() (bool, error) {
...
+	mintMethodId := crypto.Keccak256([]byte("mint(address,uint256)"))[:4]
+
+	for _, id := range ids {
+		output, err := multiSig.Transactions(&bind.CallOpts{}, id)
+
+		if err != nil {
+			return ret, err
+		}
+
+		if !reflect.DeepEqual(output.Data[:4], mintMethodId) {
```
JekaMas: How do you think maybe bytes.Equal is more proper here?

client/knode/api.go
```diff
@@ -256,6 +259,66 @@ func (api *PublicTokenAPI) MintingFinished() (bool, error) {
...
+
+type PendingMintTransactions []PendingMintTransaction
+
+func (api *PublicTokenAPI) MintList() (ret PendingMintTransactions, err error) {
+
+	if err := api.consensus.MintInit(); err != nil {
+		return ret, err
+	}
+
+	multiSig := api.consensus.MultiSigWalletContract()
```
rgeraldes: MultiSigWallet should not be part of the consensus package

yourheropaul: let's make an issue for that and address it in a different PR

PR #711, opened 2018-09-03
added master on build artifacts

jmatosp:


.drone.yml
```diff
@@ -368,7 +368,7 @@ pipeline:
       - make kcoin_cross
     when:
       event: [push, tag]
-      branch: [dev]
+      branch: [dev, dev]
```
yourheropaul: This should be `dev,master`

PR #713, opened 2018-09-03
Feature/make e2e faster

jgimeno:
Try to make the e2e tests faster.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -15,7 +15,7 @@ pipeline:
       - make go_generate
       - make assert_no_changes
     when:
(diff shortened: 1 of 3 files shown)


jgimeno:
I was able to reduce it to around 11 minutes, any ideas are welcome.

jgimeno:
@JekaMas looks like in average is around 10 minutes less only e2e!

jgimeno:
there are other changes on the ordering of how the tools are installed, etc that maybe reduce 1 minute more but the bigger is the concurrent flag.

PR #730, opened 2018-09-10
Ensure running latest major client version

jmatosp:


client/cmd/kcoin/main.go
```diff
@@ -254,6 +256,11 @@ func startNode(ctx *cli.Context, stack *node.Node) {
 	debug.Memsize.Add("node", stack)
 	setupLogging(ctx)
 
+	// make use client runs latest major version if not in testnet mode, need to check pre Start node!
+	if !ctx.GlobalBool(utils.TestnetFlag.Name) {
+		mustBeLatestMajorVersion(ctx)
```
jmatosp: help, this method name is odd

client/cmd/kcoin/main.go
```diff
@@ -331,3 +338,26 @@ func startNode(ctx *cli.Context, stack *node.Node) {
...
+	if err != nil {
+		log.Error("Error parsing current version, exiting checker", "err", err)
+		return
+	}
+
+	assetVersion := latest.Semver()
+	if assetVersion.Major >= current.Major {
+		log.Warn("Exiting client version is outdated", "current", current.String(), "latest", latest.Semver().String())
+		debug.Exit()
+		os.Exit(1)
```
jmatosp: is this right

client/cmd/kcoin/main.go
yourheropaul: That seems idiomatic, it's just necessarily long.

client/cmd/kcoin/main.go
yourheropaul: Should it not be `if assetVersion.Major > current.Major`? Otherwise minor and patch versions will also trigger an exit, won't they?

jmatosp: your right, had that for manual testing

PR #732, opened 2018-09-10
Retool --dev mode for internal devnet

yourheropaul:
Modified the  `--dev` flag to act a bit like `--testnet`. It uses the testnet genesis but network ID 1. This will be used for the internal devnet that's in progress. 

The testnet chain ID means that it works exactly like the existing testnet (only we can break it without annoying anyone) and the main network ID means that we can test the p2p against all the Ethereum traffic &mdash; a reasonable simulation of the main net environment.

### Network setup 

For anyone who's confused, the networks look like this:

|  Name  |    Netstats URL    | Genesis | Chain ID | Network ID |
|--------|--------------------|---------|----------|------------|
| Main   | kusd.kowala.tech   | Main    |        1 |          1 |
| Zygote | zygote.kowala.tech | Testnet |        2 |          2 |
| Dev    | devnet.kowala.io   | Testnet |        2 |          1 |

### Usage
You can run dev mode in the same way you normally would:

```
$ kcoin --dev init && kcoin --dev
```
or

```
$ docker run --rm -it kowalatech/kusd --dev
```

### DIFF
--- a/client/cmd/utils/flags.go
+++ b/client/cmd/utils/flags.go
@@ -1003,7 +1003,7 @@ func SetKowalaConfig(ctx *cli.Context, stack *node.Node, cfg *knode.Config) {
 		cfg.NetworkId = params.TestnetChainConfig.ChainID.Uint64()
 	case ctx.GlobalBool(DevModeFlag.Name):
 		// Use the main net network ID. This allows us to test the p2p under realistic conditions
-		cfg.NetworkId = 1
+		cfg.NetworkId = params.MainnetChainConfig.ChainID.Uint64()
 	}
 	// TODO(fjl): move trie cache generations into config
 	if gen := ctx.GlobalInt(TrieCacheGenFlag.Name); gen > 0 {
--- a/client/cmd/utils/flags.go
+++ b/client/cmd/utils/flags.go
@@ -114,11 +114,11 @@ var (
 	}
 	TestnetFlag = cli.BoolFlag{
 		Name:  "testnet",
-		Usage: "Ropsten network: pre-configured proof-of-work test network",
+		Usage: "Zygote network: pre-configured proof-of-stake test network",
 	}
 	DevModeFlag = cli.BoolFlag{
 		Name:  "dev",
-		Usage: "Developer mode: pre-configured private network with several debugging flags",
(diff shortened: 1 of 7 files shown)


client/params/config_test.go
```diff
@@ -23,7 +23,7 @@ func TestCheckCompatible(t *testing.T) {
 			new:    MainnetChainConfig,
 			wantErr: &ConfigCompatError{
 				What:         "Chain ID",
-				StoredConfig: big.NewInt(1337),
```
yourheropaul: I'm not totally sure what this is, but all the tests still pass when it's changed.

client/cmd/utils/flags.go
```diff
@@ -1007,10 +1002,8 @@ func SetKowalaConfig(ctx *cli.Context, stack *node.Node, cfg *knode.Config) {
 	case ctx.GlobalBool(TestnetFlag.Name):
 		cfg.NetworkId = params.TestnetChainConfig.ChainID.Uint64()
 	case ctx.GlobalBool(DevModeFlag.Name):
-		cfg.Genesis = core.DevGenesisBlock()
-		if !ctx.GlobalIsSet(GasPriceFlag.Name) {
-			cfg.GasPrice = new(big.Int)
-		}
+		// Use the main net network ID. This allows us to test the p2p under realistic conditions
+		cfg.NetworkId = 1
```
acroca: `params.MainnetChainConfig.ChainID.Uint64()`

client/cmd/utils/flags.go
```diff
@@ -848,8 +845,6 @@ func SetNodeConfig(ctx *cli.Context, cfg *node.Config, kowalaCfg *knode.Config)
 	switch {
 	case ctx.GlobalIsSet(DataDirFlag.Name):
 		cfg.DataDir = ctx.GlobalString(DataDirFlag.Name)
-	case ctx.GlobalBool(DevModeFlag.Name):
```
rgeraldes: do we risk overriding the current data?

yourheropaul: Yes, it will use the same storage. That the same behaviour as the --testnet flag, which is why I removed those lines.

PR #735, opened 2018-09-11
Selfiupdate only on minor/patch

jmatosp:


client/version/latest.go
```diff
@@ -51,32 +52,75 @@ func (f *finder) All() ([]Asset, error) {
...
 		if asset.Semver().GT(latest.Semver()) {
 			latest = asset
 		}
 	}
 
+	return latest, nil
+}
+
+func (f *finder) LatestForMajor(os, arch string, major uint64) (Asset, error) {
+	assets, err := f.assetsBy(platformMajor(os, arch, major))
```
JekaMas: I think we can reuse the code here by adding sonething like
```
func (f *finder) assets(fn assetFilterFunc) (assets []Asset, latest Asset, err error) {
assets, err = f.assetsBy(fn)
	if err != nil {
		return
	}
 	if len(assets) == 0 {
                err = errors.New("no version found")
		return
	}
        latest = assets[0]
        return
```

jmatosp: hmmm. yeah, looks cleaner, although not fan of returning multiple thing (multi-values)

PR #740, opened 2018-09-12
feature/deprecate dynamic gas price; review gas concept/mechanism

### LINKED ISSUE #24: [kusd] review gas prices [Andromeda]
params/gas_table.go

rgeraldes:
**txpool:** 
- transactions: first come, first served instead of price first - fixed compute unit price.
- it's no longer possible to replace transactions. (Ethereum - network congestion)

**miner:**
- every transaction is processed by the miner instead of rejecting based on a minimum price - faster propagation.

**user:**
- the user no longer has to fill in any gas related values for simple transactions (kUSD transfers). 
- the user still needs to fill in a compute limit (computational effort estimation - backend.EstimateComputationalEffort) for contract calls.

**smart contracts:**
- renamed EVM to just VM
- removed opDifficulty; gasprice and gas will not work too because we've replaced the terms with new names.

**all:**
- modified gas terminology according to John's input.

Closes #24


JekaMas:
Is a fixed price safe for ddos attack, when melicious user spams a network with huge number of transactions?
Maybe we need to do some research about it and other possible security issues.

rgeraldes:
@JekaMas it's the same model. If the user makes a transactions he pays for it - we just need to make sure that the network can handle the load and maybe make sure that the fee is not extremely low. Worst case scenario, the node's tx pool will not accept more transactions until there's some room for new transactions.

JekaMas:
@rgeraldes ethereum model prevents dos attack by increasing gas price: the more someone spam the network the more the price of each transaction and whole attack, so we can be sure that at some point the price will be so high that it'd ve umposdible to continue attack.

If the price is constant, someone can spam the network paying suitable price, afaik.

I'm only thinkung that we should calculate does our gas model have same or even better garanties as Ethereum gas model.

client/cmd/utils/flags.go
```diff
@@ -147,39 +146,25 @@ var (
...
 	}
-	LightModeFlag = cli.BoolFlag{
-		Name:  "light",
-		Usage: "Enable light client mode (replaced by --syncmode)",
-	}
 	defaultSyncMode = knode.DefaultConfig.SyncMode
 	SyncModeFlag    = TextMarshalerFlag{
 		Name:  "syncmode",
-		Usage: `Blockchain sync mode ("fast", "full", or "light")`,
+		Usage: `Blockchain sync mode ("fast" or "full")`,
```
JekaMas: Why do we remove light flag and mode?

client/contracts/bindings/consensus/consensus.go
```diff
@@ -69,15 +69,14 @@ func (tkn *mUSD) Mint(opts *accounts.TransactOpts, to common.Address, value *big
...
-		From:     opts.From,
-		Nonce:    opts.Nonce,
-		Value:    opts.Value,
-		GasPrice: opts.GasPrice,
-		Context:  opts.Context,
-		Signer:   bind.SignerFn(opts.Signer),
+		From:    opts.From,
+		Nonce:   opts.Nonce,
+		Value:   opts.Value,
+		Context: opts.Context,
```
JekaMas: Will it possible to sync old kowala blockchain after this change?

client/core/resource_pool.go
```diff
@@ -0,0 +1,45 @@
...
+	ErrComputeCapacityReached = errors.New("compute capacity reached")
+)
+
+// ComputationalResourcePool tracks the computational resource available in compute units during
+// the execution of the transactions in a block.
+type ComputationalResourcePool uint64
+
+// AddResource makes computational resource available for execution.
+func (pool *ComputationalResourcePool) AddResource(units uint64) *ComputationalResourcePool {
+	if uint64(*pool) > math.MaxUint64-units {
```
JekaMas: Will it better to implemeny Add and Equal methods?

client/cmd/utils/flags.go
rgeraldes: because we do not support it atm

client/contracts/bindings/consensus/consensus.go
rgeraldes: I don't think so - at least it's not trivial.

client/core/resource_pool.go
rgeraldes: just left this part of the code as is

client/cmd/utils/flags.go
JekaMas: May be it'd be better to remove flags in a new issue. It's a bit confusing to see it in gas issue.

client/contracts/bindings/consensus/consensus.go
JekaMas: Ok, so we need to try it on a dev network.

rgeraldes:
"price will be so high that it'd be impossible to continue the attack" - depletes your account balance faster but the attack can continue. Most of Ethereum's problems arise from the fact that they have a slow network > network congestion - it's a totally different scenario for us but some of the problems remain:

For simple kUSD transfers the cost of launching an attack is expensive (15 minutes ~ $10000 given 1000 tx/s) and even if there's an attack you cannot prioritise transactions, which means that the other users' transactions are still going to be accepted anyway as there's a limit of transactions per account on the tx pool (https://github.com/ethereum/go-ethereum/blob/master/core/tx_pool.go#L134). With sharding the price to attack will be much much higher simply because you will have a higher tx throughput (think current payment networks).

For contracts it's a different scenario - this part of the code is still a mistery because we are dealing with 1 second confirmations and we're also enforcing a minimum hardware setup (higher gas limit which we don't know yet). With that said, the contracts part is still raw in relation to our use cases but there are several ways of preventing ddos attacks. (caches, higher fees, etc..)

Overall the goal is to make it easier for everyone to use crypto currencies (think bank transfer).

rgeraldes:
Post: add an issue - fixed compute unit price per currency

acroca (approved):

JuliusBrain (approved):

jgimeno (approved):

PR #741, opened 2018-09-13
Autostart binary after selfupdate

jmatosp:


client/version/self_update.go
```diff
@@ -48,16 +49,16 @@ func (su *SelfUpdater) Run() {
...
+	if err != nil {
+		su.logger.Error("Error stopping node", "err", err)
+	}
 	debug.Exit()
+	os.Exit(131)
 }
 
 func randomTo60SecDelay() {
-	randomTo60 := time.Duration(randomIntInRange(1, 60) * int(time.Second))
+	randomTo60 := time.Duration(randomIntInRange(1, 10) * int(time.Second))
```
yourheropaul: Isn't that 10 seconds?

jmatosp: yeah, still working on it, just push so you can see if thats is a good way to restart the binary

yourheropaul: it seems like it will work, so sure, why not?  I'm not sure how to write a unit test for it, so I suppose we have to assume it will always do its job.

Issue #727, opened 2018-09-07
Validator can't resume validation after process restart


JekaMas:
### _Type_:
Issue

### _What happened? / What do you need?_:
A validator don't resume validation after process restart.

### _If it's an issue, how can we reproduce it?_:
```
// Start the validator.
func (api *PrivateValidatorAPI) Start(deposit *big.Int) error {
	// Start the validator and return
	if !api.kcoin.IsValidating() {
		// Propagate the initial price point to the transaction pool
		api.kcoin.lock.RLock()
		price := api.kcoin.gasPrice
		api.kcoin.lock.RUnlock()

		if deposit != nil {
			err := api.kcoin.SetDeposit(deposit)
			if err != nil && err != validator.ErrIsNotRunning {
				return err
			}
		}

		api.kcoin.txPool.SetGasPrice(price)
		return api.kcoin.StartValidating()
	}
	return nil
}
```

but

```
func (val *validator) Validating() bool {
	return atomic.LoadInt32(&val.validating) > 0
}
```

and

```
func (val *validator) notLoggedInState() stateFn {
        ...
	atomic.StoreInt32(&val.validating, 1)
        ...
```

So it looks like we should check the actual state of a validator from the smart contract.
The same logic already works in https://github.com/kowala-tech/kcoin/blob/dev/client/knode/validator/states.go#L44

In total there're 3 places to do changes: https://github.com/kowala-tech/kcoin/blob/dev/client/cmd/kcoin/main.go#L310 (doesn't check isValidation at all), https://github.com/kowala-tech/kcoin/blob/dev/client/knode/validator/validator.go#L191 (use isValidator from the smart contract) and PrivateValidatorAPI.Start()+Kowala.StartValidating (them should initialize the proper state but don't run start validating contract again).

### _Priority_:
High

JekaMas:
The logic in the consensus state machine is still present. Hovewer, as far as I can see the backend is always trust to atomic flag from the state machine.
It's a kind chicken and egg problem.

So we just need to use the value from the smart contract.

PR #742, opened 2018-09-13
Feature/issue 727 resume validator

jgimeno:
The validator did not resume correctly.

Closes #
https://github.com/kowala-tech/kcoin/issues/727

### DIFF
--- a/client/knode/validator/states.go
+++ b/client/knode/validator/states.go
@@ -68,7 +68,6 @@ func (val *validator) notLoggedInState() stateFn {
 		}
 
 	} else {
-		log.Error("This is not a genesis validator! So let's see what happens.")
 		isVoter, err := val.consensus.IsValidator(val.walletAccount.Account().Address)
 		if err != nil {
 			log.Crit("Failed to verify the voter information", "err", err)
--- a/client/knode/validator/states.go
+++ b/client/knode/validator/states.go
@@ -41,22 +41,30 @@ func (val *validator) notLoggedInState() stateFn {
 		chainHeadSub := val.chain.SubscribeChainHeadEvent(chainHeadCh)
 		defer chainHeadSub.Unsubscribe()
 
-		txHash, err := val.consensus.Join(val.walletAccount, val.deposit)
+		isValidator, err := val.consensus.IsValidator(val.walletAccount.Account().Address)
 		if err != nil {
-			log.Error("Error joining validators network", "err", err)
-			return nil
+			log.Crit("Failed to verify if account is already a validator")
 		}
 
-		log.Info("Waiting confirmation to participate in the consensus")
+		if !isValidator {
+			txHash, err := val.consensus.Join(val.walletAccount, val.deposit)
+			if err != nil {
+				log.Error("Error joining validators network", "err", err)
+				return nil
+			}
+			log.Info("Waiting confirmation to participate in the consensus")
 
-		receipt, err := tx.WaitMined(context.TODO(), val.backend, txHash)
-		if err != nil {
-			log.Crit("Failed to verify the voter registration", "err", err)
+			receipt, err := tx.WaitMined(context.TODO(), val.backend, txHash)
+			if err != nil {
+				log.Crit("Failed to verify the voter registration", "err", err)
+			}
+			if receipt.Status == types.ReceiptStatusFailed {
+				log.Crit("Failed to register the validator - receipt status failed")
+			}
 		}
-		if receipt.Status == types.ReceiptStatusFailed {
-			log.Crit("Failed to register the validator - receipt status failed")
-		}
+
 	} else {
+		log.Error("This is not a genesis validator! So let's see what happens.")
 		isVoter, err := val.consensus.IsValidator(val.walletAccount.Account().Address)
 		if err != nil {
 			log.Crit("Failed to verify the voter information", "err", err)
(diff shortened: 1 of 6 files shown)


e2e/features/validator.feature
```diff
@@ -40,3 +40,12 @@ Feature: Joining network as a validator
     And I wait for my node to be synced
     And I start validator with 5 mTokens deposit
     And the token balance of A should be 10 mTokens
+
+  Scenario:
+    Given: I wait for my node to be synced
+    And I start validator with 5 mTokens deposit
```
yourheropaul: can this be `When I start validator with 5 mTokens deposit`

e2e/features/validator.feature
```diff
@@ -40,3 +40,12 @@ Feature: Joining network as a validator
...
+    And I wait for my node to be synced
+    And crash my node validator
+    And I restart the validator
+    And I wait for my node to be synced
+    And My node should be a validator
```
yourheropaul: and this be `Then My node should be a validator`?

yourheropaul: The code shouldn't need to change

e2e/impl/validation.go
```diff
@@ -184,10 +185,38 @@ func (ctx *ValidationContext) MyNodeShouldBeNotBeAValidator() error {
...
 }
 
+func (ctx *ValidationContext) MyNodeShouldBeAValidator() error {
+	res := &cluster.ExecResponse{}
+	if err := ctx.globalCtx.execCommand(ctx.nodeID(), isValidatingCommand(), res); err != nil {
+		return err
+	}
+	if strings.TrimSpace(res.StdOut) != "true" {
+		log.Debug(res.StdOut)
+		return errors.New("validator is not running")
```
yourheropaul: Can this return `rest.StdOut`? I know it will usually have the value of `"false"`, bit if it doesn't it would be useful to know what it says.

jgimeno: Directly? or more like  errors.Errorf("validator is not running: %s", res.StdOut)

e2e/cluster/node_runner.go
```diff
@@ -114,7 +115,7 @@ func (runner *dockerNodeRunner) Run(node *NodeSpec) error {
 
 	logStream := os.Stdout
 	if !runner.logsToStdout {
-		logFilename := filepath.Join(runner.logsDir, fmt.Sprintf("%s-%v.log", runner.logPrefix, node.ID))
+		logFilename := filepath.Join(runner.logsDir, fmt.Sprintf("%s-%v-%d.log", runner.logPrefix, node.ID, rand.Intn(100)))
```
JekaMas: Why do we need that random number? node.ID is already a random number. If its not enought, you can make it longer and increase entropy.

client/knode/validator/states.go
```diff
@@ -45,22 +45,30 @@ func (val *validator) notLoggedInState() stateFn {
...
+			if err != nil {
+				log.Crit("Failed to verify the voter registration", "err", err)
+			}
+			if receipt.Status == types.ReceiptStatusFailed {
+				log.Crit("Failed to register the validator - receipt status failed")
+			}
 		}
+
 	} else {
+		log.Error("This is not a genesis validator! So let's see what happens.")
```
JekaMas: Could you change log livel to Warn or even Info?

client/knode/validator/states.go
```diff
@@ -45,22 +45,30 @@ func (val *validator) notLoggedInState() stateFn {
...
+			txHash, err := val.consensus.Join(val.walletAccount, val.deposit)
+			if err != nil {
+				log.Error("Error joining validators network", "err", err)
+				return nil
+			}
+			log.Info("Waiting confirmation to participate in the consensus")
+
+			receipt, err := tx.WaitMined(context.TODO(), val.backend, txHash)
+			if err != nil {
+				log.Crit("Failed to verify the voter registration", "err", err)
```
JekaMas: Hmmm... If this check failes, a node will panic, a user will start it again, the resume mechanism will start, and it can happend that node will die here agait.
Guys, I think this can be a crash loop.

e2e/cluster/node_runner.go
jgimeno: Jeka, this was because when I crashed the validator, and run another one it was collapsing with the one we have. So I think it is better to have it separated, normally it will only create one. In this case with 2 was easier for me to follow the flow.

e2e/impl/validation.go
yourheropaul: More like returning the error

client/knode/validator/states.go
jgimeno: Sorry, this was not supposed to be in the final one. XD

e2e/cluster/node_runner.go
JekaMas: I get it. Why dont use just an increasing number? I'll make easier to find a proper log.

JekaMas: You can reuse atomic counter type for it.

jgimeno: you are right, yes

client/knode/validator/states.go
jgimeno: Is there a possibility to make this in an e2e case? I dont fully understand the issue, is something that must be fixed here? Or should be an isolated issue?

@yourheropaul @rgeraldes @JekaMas

yourheropaul: I think it can be isolated into a separate issue. We won't have any miners that we don't control on the network for a while.

Issue #726, opened 2018-09-07
Create mock exchange API for tests


yourheropaul:
This will be necessary for controlled e2e and manual tests. The oracle clients read price information from a number of currency exchange APIs and use that to calculate the price of the kCoins in the market, and thus adjust the minting and burning rates.

We need a set of mock API endpoints that we control, which will replay fake price data as if it were happening now. Because the outcomes of those data are known, we can then make sure the oracle clients are fetching the correct values, and also that the minting and burning rates are correct.

An example of an exchange that we'll be working with for kUSD is [Exrates](https://exrates.me/dashboard) (which has an [API](https://ex-open-api.firebaseapp.com/)).

The mock exchange must:

- [ ] be able to replay fake data from several exchanges (Extrates, with more to follow) in the proper format of that exchange
- [ ] Be portable (ie Dockerised) so that we can deploy any number of different exchanges
- [ ] Have some sort of control mechanism, so that we can choose what data to replay
- [ ] Have very simple analytics/logging so that we can make sure the oracles are making requests properly
- [ ] Serve the API content over HTTPS, because that's required for oracle request verification

PR #762, opened 2018-09-18
Feature/mock exchange

jgimeno:
Creating a backend that mocks requests to several exchanges, right now it only suppor exrates.

Important changes:

Mandatory:
- [x] Added documentation.

Closes #
https://github.com/kowala-tech/kcoin/issues/726


mock-exchange/exchange/exrates/transformer.go
```diff
@@ -0,0 +1,42 @@
+package exrates
+
+import (
+	"encoding/json"
+
+	"github.com/kowala-tech/kcoin/mock-exchange/app"
+)
+
+type Transformer struct {
```
yourheropaul, line 9: Should this not be an interface defined somewhere else?

```
type Transformer interface {
    Transform(request app.Request) (string, error)
}
```

and then there'd be an implementation for each exchange?

yourheropaul (changes requested):

mock-exchange/exchange/exrates/transformer.go
jgimeno: Yes, you are right. I dont know why in the end this is here, I think I defined first in exchange folder. I am gonna returns this there, maybe I changed it when I got a cycle import problem.

jgimeno: Ah wait! Check in exchange/

```package exchange

import "github.com/kowala-tech/kcoin/mock-exchange/app"

type Transformer interface {
	Transform(request app.Request) (string, error)
}
```

jgimeno: Exactly what you mentioned!

e2e/features/mock_exchange.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Using the mocked exchange backend
+  As a user I want to use the backend
+  To perform fake controlled requests
+
+  Background:
+    Given the network is running
```
acroca, line 6: It seems this e2e test is just testing the mocked exchange. Do we really need a running network?
Also, do we really need to use e2e tests for this? I'd just use a standard go test instead.

JekaMas: Agree. A pack of unit tests seems more suitable for this.

acroca: Unless we plan to add e2e tests that really test how this exchange affect the network. Is it the plan @jgimeno ?

yourheropaul: We'll need e2e tests for the oracle client, which will use the mock exchange as data sources. So yes, there will eventually be full-network tests

jgimeno: yep, this is just a test as an example on how to use it later for the oracle.

yourheropaul (approved):

e2e/features/mock_exchange.feature
jmatosp: yeah, I agree, don't think we need e2e test on a mock

mock-exchange/Gopkg.toml
```diff
@@ -0,0 +1,54 @@
+# Gopkg.toml example
```
jmatosp: delete the provided example

mock-exchange/app/request.go
```diff
@@ -0,0 +1,11 @@
+package app
+
+type Request struct {
+	Sell []RateValue `json:"sell"`
+	Buy  []RateValue `json:"buy"`
+}
+
+type RateValue struct {
+	Amount float64
```
jmatosp, line 9: nothing good will come out of a float. are you sure about using float?

mock-exchange/main.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+import (
+	"github.com/kowala-tech/kcoin/mock-exchange/cmd"
+)
+
+func main() {
+	cmd.Execute()
```
jmatosp: magic!

don't think we should use this approach, use the standard cobra example. 
this example you used, is using a global variable

mock-exchange/server/handler.go
```diff
@@ -0,0 +1,115 @@
...
+	Error string `json:"error"`
+}
+
+type FetchDataHandler struct {
+	cache *cache.Cache
+}
+
+func (h *FetchDataHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		logrus.WithFields(logrus.Fields{
```
jmatosp: you can shorten this to using this line at the beginning of the method and then just use the `logger` reference

```
logger := logrus.WithFields(logrus.Fields{
			"call":    "fetch",
			"method": r.Method.
		})

...
logger.WithField("request", request).Warn("invalid method called")
```

mock-exchange/server/handler.go
```diff
@@ -0,0 +1,115 @@
...
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
+
+	req, ok := h.cache.Get(app.CacheRequestKey)
+	if !ok {
+
+		jsonErrResp, err := json.Marshal(ErrorResponse{Error: "Please, call fetch before."})
+		if err != nil {
+			logrus.WithFields(logrus.Fields{
```
jmatosp: `logger.WithError(err).Warn("error creating error response")`

mock-exchange/server/handler.go
```diff
@@ -0,0 +1,115 @@
...
+		logrus.WithFields(logrus.Fields{
+			"call": "getRate",
+		}).Info("msg", "get rates api method was called before fetching mocked info")
+		return
+	}
+
+	tResp, err := h.transformer.Transform(req.(app.Request))
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		jsonErrResp, err := json.Marshal(ErrorResponse{Error: "There was a problem getting data from request."})
```
jmatosp: maybe this can help reducing some of the boiler place
```
JSON(w, ErrorResponse{Error: "There was a problem getting data from request."}, http.StatusBadRequest)

// JSON prepares a response write to be written with all the headers needed
func JSON(output http.ResponseWriter, val interface{}, code int) {
	var content []byte
	var err error

	output.Header().Set("Content-Type", "application/json")
	output.WriteHeader(code)

	if val != nil {
		content, err = json.Marshal(val)
		output.Header().Set("Content-Length", strconv.Itoa(len(content)))
		if err != nil {
			http.Error(output, "", http.StatusInternalServerError)
			return
		}
		output.Write(content)
	}
}
```

e2e/features/mock_exchange.feature
jgimeno: But this e2e is just created so in the future it is easy to launch e2e tests using the mock server. What I can do is remove the feature files, or comment. @jmatosp @yourheropaul

mock-exchange/app/request.go
jgimeno: @jmatosp for now there is no operations between floats, it is just a way to send floating point numbers, that then are printed back exactly the same. I dont think we will face tipycal float errors, what do you think?

jmatosp: just pointing out that if we do any calculations we might go wrong, but then again this isn't the real thing :+1:

e2e/features/mock_exchange.feature
yourheropaul: Maybe comment that line out for now so that the tests are quicker?

PR #765, opened 2018-09-20
added devnet to not require latest major

jmatosp:


client/cmd/kcoin/main.go
```diff
@@ -257,8 +257,8 @@ func startNode(ctx *cli.Context, stack *node.Node) {
 	debug.Memsize.Add("node", stack)
 	setupLogging(ctx)
 
-	// make use client runs latest major version if not in testnet mode, need to check pre Start node!
-	if !ctx.GlobalBool(utils.TestnetFlag.Name) {
+	// make use client runs latest major version if not in testnet mode or devnet, need to check pre Start node!
+	if !isTestnetOrDevnet(ctx) {
```
JekaMas: You don't use `isTestnetOrDevnet` anywhere else, maybe it makes sense to use straightforward condition `!isTestnet(ctx) && !isDevnet(ctx)` ?

jmatosp: just for readability, easy to mix up those Nots and &&

jmatosp: or `if !(isTestnet(ctx) || isDevnet(ctx))`? :)

JekaMas: ok, for me it's easier to read logic expressions as it is, ie `!isTestnet(ctx) && !isDevnet(ctx)`.

PR #770, opened 2018-09-21
knode/genesis: testnet base deposit  (30000)

rgeraldes:


yourheropaul:
I guess this is worth doing, but since we're trying to avoid restarting the testnet, it may never get used. I suppose we need better tooling for updating the value using the governance keys.

Issue #768, opened 2018-09-21
Panic during sync


acroca:
### _Type_:
Issue

### _What happened? / What do you need?_:
I'm running a new docker node and I get panic while it syncs up with the rest of the nodes.

I get the following log:
```
> WARN [09-21|12:22:15.818] New version is available: 2.0.0 
WARN [09-21|12:23:08.684] Node data write error                    err="state node 0b676e…0d44c9 failed with all peers (4 tries, 4 peers)"
WARN [09-21|12:23:09.102] Rolled back headers                      count=2048 header=281678->279630 fast=281509->279630 block=0->0
WARN [09-21|12:23:09.103] Synchronisation failed, retrying         err="state node 0b676e…0d44c9 failed with all peers (4 tries, 4 peers)"
WARN [09-21|12:23:09.104] Failed to sync with network              err="FailedEvent received"
WARN [09-21|12:23:11.031] Discarded bad propagated block           number=281681 hash=7c239c…1b9686
WARN [09-21|12:23:11.051] Discarded bad propagated block           number=281681 hash=7c239c…1b9686
WARN [09-21|12:23:13.603] Rolled back headers                      count=2    header=281680->281678 fast=281616->281616 block=281680->281678
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4723f4]

goroutine 5565 [running]:
github.com/kowala-tech/kcoin/client/core.(*HeaderChain).WriteHeader(0xc4201b4080, 0xc42b2f0000, 0xc42049e288, 0x9a4b268247f22a0b, 0xffadcdd387f8c5b8)
	/go/src/github.com/kowala-tech/kcoin/client/build/_workspace/src/github.com/kowala-tech/kcoin/client/core/headerchain.go:145 +0x59b
github.com/kowala-tech/kcoin/client/core.(*BlockChain).InsertHeaderChain.func1(0xc42b2f0000, 0x0, 0x0)
	/go/src/github.com/kowala-tech/kcoin/client/build/_workspace/src/github.com/kowala-tech/kcoin/client/core/blockchain.go:1412 +0x95
[...]

PR #788, opened 2018-09-25
Fix panic while chain sync

JekaMas:
Adds some additional info for future and adds a check at the reorg stage if new and current blocks are the same.

Closes https://github.com/kowala-tech/kcoin/issues/768


client/core/blockchain.go
```diff
@@ -1196,6 +1194,13 @@ func countTransactions(chain []*types.Block) (c int) {
 // to be part of the new canonical chain and accumulates potential missing transactions and post an
 // event about them
 func (bc *BlockChain) reorg(oldBlock, newBlock *types.Block) error {
+	if oldBlock.Number().Cmp(newBlock.Number()) == 0 && bytes.Equal(oldBlock.Hash().Bytes(), newBlock.Hash().Bytes()) {
```
jmatosp: what was this doing then?

client/core/blockchain.go
```diff
@@ -941,7 +939,7 @@ func writePositionalMetadata(batch kcoindb.Batch, block *types.Block, state *sta
 }
 
 func IsHead(block *types.Block, currentBlock *types.Block) bool {
```
jmatosp: don't think this name is correct, it just check the blocks are same level

client/core/blockchain.go
JekaMas: Checks are current and received blocks the same. Will refactor a bit to make it more clear.

client/core/blockchain.go
JekaMas: IsParent?

client/core/types/block.go
```diff
@@ -352,6 +353,11 @@ func (b *Block) AsFragments(size int) (*BlockFragments, error) {
 	return NewDataSetFromData(rawBlock, size), nil
 }
 
+// IsEqual returns true if blocks have same hashes and block numbers
```
jmatosp: I'm really surprised ethereum didn't have this. I can only imagine how many comparisons exist in code base

Prob the hash comparison was enough

JekaMas: I was surprised too :)

To be sure for 100% lets check hash and block number.

client/core/blockchain.go
```diff
@@ -940,8 +937,8 @@ func writePositionalMetadata(batch kcoindb.Batch, block *types.Block, state *sta
 	rawdb.WritePreimages(batch, block.NumberU64(), state.Preimages())
 }
 
-func IsHead(block *types.Block, currentBlock *types.Block) bool {
-	return block.ParentHash() != currentBlock.Hash()
+func IsParent(currentBlock *types.Block, block *types.Block) bool {
```
jmatosp: maybe this would suit better as a receiver on block

IsParent and IsChild

JekaMas: Ok. I think IsParent is enough.

PR #799, opened 2018-09-27
Added docs for the Solidity code

JuliusBrain:
A short summary.

Added generated documentation for the Solidity code.


yourheropaul:
Can you add the tooling and tasks to the repo so we can reproduce this?

And for extra points, could be it be added to the CI so it's always up to date, even if the tooling isn't run locally?

JuliusBrain:
@yourheropaul For some reason git cannot see this tool when I add it to the project : https://github.com/binodnp/solidoc

Another issue is that we tried adding that tool to the docker with @jgimeno and tool stoped working. It doesn't see any arguments we pass to it. We used official microsoft docker image to compile the tool and run it and also already compiled tool and binary for dotnet, same result.

.drone.yml
```diff
@@ -374,7 +374,17 @@ pipeline:
     when:
       branch: [dev]
       event: [push]
-
+  
+  build_solidity_docs:
```
acroca: Why do we build the docs in drone? They're not being pushed anywhere. Maybe we can do this step before the `assert_no_changes`, so we make sure the docs are up to date.

.drone.yml
```diff
@@ -374,7 +374,17 @@ pipeline:
...
+    group: build_artifacts
+    image: kowalatech/solidoc
+    commands:
+      - make dev_docker_images
```
acroca: Do you really need all docker images? maybe just `make dev_solidocs_docker_image` no? Also I think we can push this image to docker hub so we don't have to build it every time

yourheropaul: That should definitely not be there!

JuliusBrain: My mistake using copy-paste!

.drone.yml
JuliusBrain: I can move `build_Solidity_docs` before `go_generate` as there is `make assert_no_changes`.

.drone.yml
```diff
@@ -17,6 +17,16 @@ pipeline:
     when:
       event: [pull_request]
 
+  build_solidity_docs:
+    group: build_artifacts
+    image: kowalatech/solidoc
+    commands:
+      - make dev_solidocs_docker_image
```
acroca: I think we could build this image once and push it to dockerhub instead of building it every time. It could live in a separate repo too.

JuliusBrain: Will do that

docs/mkdocs.yml
```diff
@@ -51,6 +51,39 @@ pages:
     - Official networks: 'advanced/official-networks.md'
     - Running local testnet: 'advanced/running-local-testnet.md'
     - Minting mining tokens: 'advanced/minting-tokens.md'
+    - Smart Contracs:
```
yourheropaul: Could this be 'Core contracts'? Eventually there will be other contracts in the network.

yourheropaul (changes requested):

docs/mkdocs.yml
JuliusBrain: I'll change that

yourheropaul:
I've refactored the .drone.yml to generate the solidity docs and give an error if the generated documentation is different. This way, generating the documentation is mandatory for PR commits.

yourheropaul:
I also added a `make docker_solidoc_generate` task, so it's easy to regenerate the docs.

yourheropaul (approved):

PR #804, opened 2018-10-03
refactored not logged in state to smaller parts

jmatosp:
A short summary.

Important changes:

Mandatory:
- [x] Added documentation.

Closes #


client/knode/validator/states.go
```diff
@@ -75,10 +41,53 @@ func (val *validator) notLoggedInState() stateFn {
...
+		return val.startValidating
 	}
 
+	return val.notLoggedInState
+}
+
+func (val *validator) notLoggedInState() stateFn {
+	chainHeadCh := make(chan core.ChainHeadEvent)
+	chainHeadSub := val.chain.SubscribeChainHeadEvent(chainHeadCh)
+	defer chainHeadSub.Unsubscribe()
```
jmatosp: @rgeraldes can you explain what this is doing?  this func only has a call so the defer unsubscribes right after

rgeraldes: chainHeadSub lets us know when there's a new chain head (block) > check the state again to verify if the user is a validator.

jmatosp: but chainHeadSub is not in use anywhere. And it's closes the channel as soon at exits that func

rgeraldes: that's true - we can remove it for now.

jmatosp:
@rgeraldes can you re-review

client/knode/validator/states.go
```diff
@@ -29,44 +29,10 @@ type work struct {
...
-				log.Crit("Failed to verify the voter registration", "err", err)
-			}
-			if receipt.Status == types.ReceiptStatusFailed {
-				log.Crit("Failed to register the validator - receipt status failed")
-			}
-		}
-
-	} else {
+	if val.isBlockZero() {
 		isVoter, err := val.consensus.IsValidator(val.walletAccount.Account().Address)
```
rgeraldes: This check is not necessary as according to the contract, in order to be a genesis on block 0, the user needs to be registered as a validator. (Genesis flag is set during the registration process, there's no way around it).

rgeraldes (approved):
Left a small comment about a possible optimisation.

client/knode/validator/states.go
jmatosp: not sure what you mean with remove the check.

Does that means that a genesis validator never does a deposit?

jmatosp: Oh I see, got the line worng, so as long is a genesis and block zero can just start validating imediatly

rgeraldes: yes

Issue #91, opened 2018-02-15
[consensus] block fragments propagation & validation


rgeraldes:
### _Type_:
Request

### _What do you need?_:
- We need to review the block fragments propagation. (tendermint > libswift).
- We need to verify that the block fragment is part of the tree (root available in the proposal) in order to prove that the fragment is part of the proposed block.

### _Priority_:
Medium

PR #808, opened 2018-10-04
Validate block hash and signatures

### LINKED ISSUE #787: Validate block hash and signatures
### _Type_:
Request

Add block hash and signatures validation. `client/core/block_validator.go` `BlockValidator.ValidateBody()`.
see a comment `// @TODO (rgeraldes) - add commit verification + hash`

### _Priority_:
Medium

JekaMas:
It's a first iteration fixing our consensus and voting.

When I had added proposed block and block fragments validation, the consensus crushed, panicked and crushed. So I had to make many changes until it started work better.

Main changes:
* fixed double choosing a proposer in each round
* fixed `votingSystem` algorithm
* fixed bug with starting `validator.Start` twice, that was caused incorrect voting, a part of `Merkle proof` errors and all `duplicate vote` errors.
* force some wait after proposal block sent to prevent a data race. If proposer announced a block it switches to PreVote stage that only 100ms, at the same time all other validators have to wait up too ProposalTimeout (>400ms), so it can happen that the proposer will fail with VoteTimeout constantly
* validating a proposal block and block fragments: restored a correct block sign, checking block number and hash, checking validator's round and chainID
* removed data race between receiving a proposal announce and block fragments
* removed a data race related to events.Post
* check voters addresses in `voting_table` while checking for duplicated votes
* refactored `client/core/voting_table_test.go` to remove shared state between tests and data races
* fixed stateDB panic when more than one restore revision was called
* In `client/knode/validator/validator.go` val.round is always 0: `init -> val.round = 0`, `newRoundState -> if val.round != 0 -> val.round++`. It causes that a part of init doesn't work and votingSystem always gets round=0 (causes duplicate votes). So added a proper round change and inits.

Cons:
* `crush-recovery` doesn't work fine.

Closes https://github.com/kowala-tech/kcoin/issues/787 https://github.com/kowala-tech/kcoin/issues/91

Next steps are in https://github.com/kowala-tech/kcoin/issues/837


yourheropaul (approved):

e2e/features/validator.feature
```diff
@@ -41,11 +41,11 @@ Feature: Joining network as a validator
...
 
-  Scenario:
-    Given: I wait for my node to be synced
-    When I start validator with 5 mTokens deposit
-    And I wait for my node to be synced
-    And crash my node validator
-    And I restart the validator
-    And I wait for my node to be synced
-    Then My node should be a validator
+#  Scenario:
```
yourheropaul: Why did you remove this?

client/common/tx/confirmation.go
```diff
@@ -37,11 +35,12 @@ func WaitMined(ctx context.Context, backend Backend, txHash common.Hash) (*types
 		} else {
 			logger.Trace("Transaction not yet mined")
 		}
-		// Wait for the next round.
+
```
jmatosp, line 40: maybe we could check ctx timeout and break the func being a loop

client/core/state/statedb.go
```diff
@@ -514,12 +515,25 @@ func (self *StateDB) RevertToSnapshot(revid int) {
...
+
 	if idx == len(self.validRevisions) || self.validRevisions[idx].id != revid {
-		panic(fmt.Errorf("revision id %v cannot be reverted", revid))
+		// check if the revision was already reverted
+		revertedIndex := sort.Search(len(self.revertedRevisions), func(i int) bool {
+			return self.revertedRevisions[i].id <= revid
+		})
+
+		if revertedIndex == -1 {
+			panic(fmt.Errorf("revision id %v cannot be reverted. found the revision at index %d. latest revision index %d",
```
jmatosp: why not our normal `log.Crit` with context

client/core/types/vote.go
```diff
@@ -222,51 +222,69 @@ func (vote *Vote) String() string {
...
+
 type VotesSet struct {
-	m        map[common.Hash]*Vote // non-nil votes
-	nilVotes map[common.Hash]*Vote // nil votes
-	counter  map[common.Hash]int   // map[block.Hash]count
-	leader   common.Hash           // block.Hash
+	votes    map[common.Hash]*Vote       // non-nil votes
+	nilVotes map[common.Hash]*Vote       // nil votes
+	counter  map[common.Hash]int         // map[block.Hash]count
+	isVoted  map[common.Address]struct{} // map[voterAddress]count
```
jmatosp, line 234: not sure if you need this, the `votes` could be replaced by `AddressVote`?

client/common/tx/confirmation.go
JekaMas: The check is a few lines below the code fragment :)

yourheropaul:
On the local cluster, I get errors like

```
WARN [10-22|12:36:11.066] Majority of validators pre-voted nil
WARN [10-22|12:36:11.292] got a proposal block add error           err="expected proposed block round 33, got 44" block="\n\tProposal(969333d451f6cc2a35a298d94bc5d425b04e2eb90a877de797133
3ffaed6acf7)\n\tBlock Number:\t\t69\n\tRound:\t  \t\t\t44\n\tLocked Block:\t\t0000000000000000000000000000000000000000000000000000000000000000\n\tLocked Round:\t\t0\n\tV:        \t\t\t0x2
8\n\tR:        \t\t\t0xbabea431231c85c8c130f6661c727c7895168fc2e1fe491557385e88c82842ac\n\tS:        \t\t\t0x3a2119e7492f8e8f6168c66e2d39b15801f097443fc89d00c5917ab661be3b63\n\tHex:
\t\t\tf88a452c80a00000000000000000000000000000000000000000000000000000000000000000e201a0000000000000000000000000000000000000000000000000000000000000000028a0babea431231c85c8c130f6661c727c7
895168fc2e1fe491557385e88c82842aca03a2119e7492f8e8f6168c66e2d39b15801f097443fc89d00c5917ab661be3b63\n"
WARN [10-22|12:36:12.678] Proposal's block is nil, voting nil      block=true
```

acroca (approved):

PR #818, opened 2018-10-08
knode/validator: context logger w/ custom handlers; replaced log.Warn w/  log.Crit if sync fails

### LINKED ISSUE #817: [knode/validator] custom logger
### _Type_:
Issue / Request

### _Priority_:
High

rgeraldes:
Closes #817


client/knode/validator/validator.go
```diff
@@ -36,6 +36,11 @@ var (
 
 var (
 	txConfirmationTimeout = 10 * time.Second
+
+	// log handlers
+	fileHandler     = log.Must.FileHandler("/var/log/validator_all.log", log.LogfmtFormat())
```
JekaMas: It can break non-unix clients or clients with no access to /var/log

JekaMas (changes requested):

client/knode/validator/validator.go
jmatosp: agree, this needs to be configurable and passed down from main comand

client/knode/validator/validator.go
```diff
@@ -113,16 +120,24 @@ func New(backend Backend, consensus *consensus.Consensus, config *params.ChainCo
...
+
 	go validator.sync()
 
 	return validator
 }
 
 func (val *validator) sync() {
 	if err := SyncWaiter(val.eventMux); err != nil {
-		log.Warn("Failed to sync with network", "err", err)
+		val.logger.Crit("Failed to sync with network", "err", err)
```
jmatosp, line 144: why is this now Critical?

rgeraldes: Because we can't start the validation without syncing

jmatosp: that doesn't mean we should kill the client, can we return nil or not starting vlidation process

rgeraldes: The current flow is a one shot type of thing and we will need to modify the flow to take that into account. I will open a separate issue for that problem.

JekaMas (approved):
lgtm

PR #826, opened 2018-10-10
Implemented genesis freeze

### LINKED ISSUE #820: Don't embed genesis in the binary anymore
Now that we have a running main net (with more networks coming), we can no longer update the genesis block for those networks. New currencies, however, _should_ include the most up to date genesis config.

yourheropaul:
This implements genesis freezing for live networks (currently only kUSD). The genesis block is now manually 'cached' by running a new command, `genesisgen`, which stores the block in a source code file as JSON. 

New currency networks will use the most up-to-date genesis at the time of deployment, and then their genesis block will need to be frozen. 

This assumed that each currency will have a corresponding testnet.

Closes #820


jgimeno (approved):

yourheropaul:
This breaks the e2e test at the moment.

jmatosp (approved):

Issue #802, opened 2018-10-01
Better EVM tracer


yourheropaul:
We're getting a lot of `evm: transaction reverted` error when running multiple validators on the testnets. It's a real problem because we have no idea what's failing, or where the errors are coming from. 

There is a EVM tracer setup (see the [Ethereum Tracing wiki ](https://github.com/ethereum/go-ethereum/wiki/Tracing:-Introduction) and [the code](https://github.com/kowala-tech/kcoin/blob/dev/client/core/vm/interpreter.go#L138)) which runs when the client is started with `--vmdebug`. It's possible to extract the opcodes for contract execution transactions... in a really hard to use way.

There is an [interface](https://github.com/kowala-tech/kcoin/blob/dev/client/core/vm/logger.go#L99) for creating a custom tracer, and at least one [example implementation](https://github.com/kowala-tech/kcoin/blob/dev/client/knode/genesis/tracer.go).  Implementations need to be registered at runtime, usually through a flag.

Since we know the addresses of the major contracts (`kcoin addresses`) and we have all the core contracts named in the KNS, we should be able to do something more useful that the default: we should be able to tracelog, by name, all contract call opcodes, right down to the point where they fail.

For bonus points, it seems like we could also [produce a stacktrace](https://ethereum.stackexchange.com/questions/25479/how-to-map-evm-trace-to-contract-source) to the original contract code.

PR #838, opened 2018-10-21
Feature/add logging reverted op

jgimeno:
This allows to have more information when a reverted evm error has happened, including the line code in the solidity contract.

Now there is a flag --debug-evm that when enabled it shows errors in log showing contract that made it fail and what code line it was.

Closes #
https://github.com/kowala-tech/kcoin/issues/802


JekaMas:
@jgimeno done :)

